PR #1140, opened 2017-10-20
Remove Tether/drop and replace with PopperJS

### LINKED ISSUE #951: Remove or replace Tether / Drop
Tether / Drop is old and hasn't been updated in two years. Recently, a team tried to render a tootlip on the server. Shallow renders don't work with Tether / Drop either (for instance in Jest testing).

Testing:
- Needs to work in Modals (z-index, render in inside `portalMount`)
- Needs to be able to flip based on viewport.
- Needs to render in Snapshots and server (no DOM environments) 

See `Dialog` component for more details.

This might be a temporary solution: https://github.com/HubSpot/drop/issues/155
This React port https://github.com/souporserious/react-popper of https://github.com/FezVrasta/popper.js might be a pretty robust solution.

Stale inactive WIP branch with partial implementation: https://github.com/salesforce-ux/design-system-react/pull/1022

interactivellama:
Fixes #951

Currently PR'd only for transparency. Tests still need to be fixed. Manual testing is working. There is a lot foundation moving around under the house for dialog using components.

### Benefits of this update:
* Use refs only (no ReactDOM.FindDOMNodes) in `<Dialog>`. Portals use `ReactDOM`, but using portals should be the exception.
* Add Portal functionality back (but not as default)
* Create new concept of position: 
    * `relative`: all CSS, no JS variant
    * `absolute`: Uses PopperJS inline variant
    * `overflowBoundaryElement`: Use a Portal with PopperJS
* Breaks Portal out into it's own component to make it close to an HOC
* Convert `isInline` usage to `position: relative` in backwards compatible way
* Expose `menuPosition` and similar props, so app dev and fix issues if non-default is needed. This would typically happen with either needing a clean HTML snapshot ( no style, position: relative or it's within a boundary parent (overflow/scroll))
* Convert `flippable` to `hasStaticAlignment` (flippable is now default across all dialogs)
* remove `constrainToScrollParent` prop (this is the default now, position absolute)
* remove flippable (backward compatible with `hasStaticAlignment`)
* Add the Storybook action add-on. `storybook/addon-actions`. The old one is deprecated
* Removes `modal` deprecation warning. Deprecation was in 0.6.x
* This library is now Jest snapshot (and Server Side Render via `ReactDOMServer.renderToStaticMarkup`) compatible with all three position modes, portal markup won't be included, but will not crash Jest like it does now.

Although there should be no breaking change as far as props go. I will plan on releasing this as a 0.8.x due to all the portal/render changes having and the possible UX behavior changes.

### Added Overlay element positioning UX pattern guidelines to document the current behavior and for future reference

 * A Dialog is content that is separate from the typical flow of a page. It typically overlays other elements in the document flow. This is achieved with elevation (`z-index`) and one of the following: relative position, absolute position, or a new top-level React render tree (portal). A boundary element is a scrolling ancestor element or the edge of the browser (window/viewport). This element typically has an overflow (overflow-y/overflow-x) style that is scroll, hidden, or auto. Inverted placement is the flipping of the overlay element from top to bottom or left to right in order stay within a boundary element.
* Dropdown menu (Combobox, DatePicker, et al.) placement is typically bottom-aligned and should be allowed to invert its placement when inside a boundary element this often happens within a modal. Dropdowns should not overflow boundary elements , since most boundary elements scroll vertically and have space for the menu.
 * If needed, left and right placed overlay elements (such as Popover and Tooltip) should be placed within a portal element attached to the DOM <body>, but styled to align to its target/trigger. Since scrolling typically occurs on the vertical axis, this allows them to overflow boundary elements and still allow scrolling of content, yet still invert placement for the browser viewport. Portal elements are only necessary if an original ancestor boundary element exists. **I'm unsure about this UX behavior and may remove it.**
* Nubbins/arrows should be repositioned for any new placement/alignment.
* Allow Overflowing of Boundary Element: Allow applications to create a portal element attached to the DOM <body> to be outside of boundary elements if manual testing shows confusing alignment/poor usability/readability. This should be exception and not the default.
* Allow disabling of inverted placement: Allow applications to prevent inverted placement if manual testing shows confusing alignment/poor usability/readability.


interactivellama:
## How this new Dialog component works

* There is no longer an inline render within components. All overlays should go through `Dialog`. `position: relative` just passes on the markup with some additional event listeners. No positional library is used.
* The default `position: absolute` will run through three renders. It will first render the overlay at `0px/0px` offset of its parent. It will then store the target DOM node, once the DOM node is set, a state change will occur and tell a PopperJS instance to be created on update, once it is created, a third render is done to update the styles created by PopperJS.
* `position: overflowBoundaryElement` will do the same three renders as `absolute` except that the initial render will create a disconnected render tree (portal) on the `body`. Then, the position will change once the target is stored.

interactivellama:
@donnieberg This is ready to be reviewed. Apologies on size! I've got some other eyes on it, too, for clarity, so please don't merge.

interactivellama:
I'll be squashing this, so I only have one commit.

components/combobox/private/menu.jsx
```diff
@@ -20,9 +20,13 @@ const propTypes = {
...
 	activeOptionIndex: PropTypes.number,
 	/**
-	 * CSS classes to be added to tag with `.slds-dropdown`. Uses `classNames` [API](https://github.com/JedWatson/classnames).
+	 * CSS classes to be added to container `div` tag. Uses `classNames` [API](https://github.com/JedWatson/classnames).
 	 */
 	className: PropTypes.oneOfType([PropTypes.array, PropTypes.object, PropTypes.string]),
+	/**
+	 * CSS classes to be added to tag with `.slds-dropdown`. Uses `classNames` [API](https://github.com/JedWatson/classnames).
+	 */
+	classNameMenu: PropTypes.oneOfType([PropTypes.array, PropTypes.object, PropTypes.string]),
```
donnieberg: I don't love that we're going against the pattern of the parent container getting`containerClassName` and its child getting `className` but I understand that you don't want to break things. I think I'd prefer having a standard pattern across all components but what are your thoughts?

interactivellama: This is private. It can break. This may have just been an oversight. I'll update.

components/lookup/lookup.jsx
```diff
@@ -87,7 +85,7 @@ const Lookup = createReactClass({
 		 */
 		filterWith: PropTypes.func,
 		/**
-		 * If true, the menu is constrained to the window and may be flipped up. Has no effect if `isInline` is `true`.
+		 * If true, the menu is constrained to the window and may be flipped up. Has no effect if `isInline` is `true`. In other components, its opposite is used `hasStaticAlignment`.
```
donnieberg: Do we need to change the description to include `menuPosition="relative" vs. isInline`?

interactivellama: Lookup still uses the old multiple sub-render `isInline` prop. I'm hesitant to change anything on a deprecated component. What do you think?

interactivellama:
I did some more performance testing. For the sake of transparency:

Combobox menus with 900 items that have SVG icons (no don't do this in your app, please. It'll be bad UX!)
* With PopperJS 760ms to no icon render and 100ms to full icon render.
* With no PopperJS 790ms to no icon render and 85ms to full icon render.
* There's appears to be an insignificant difference between use of PopperJS and not using it.
* Old TetherJS portal render is 728ms to no icon render and 108ms to full icon render. This seems to support that loading from localhost (no cache) and rendering 900 SVGs takes some time).   
* _However_, the new portal renders (because it uses refs and then re-renders twice, because you have to find the target and the portal node before you can render the contents into it) is 4x in render speed 3200ms. ReactDOM and findDOMNode have been removed, because those break abstract DOMless testing. FYI: The difference is not noticeable with 10 items which is the most an autocomplete combobox should have and portals are no longer the default in this library. Once internal consumers can move to React 16, [Portals](https://reactjs.org/docs/portals.html) will become a native thing and I'll remove the multi-render hackerary. Portals should only be used when you are doing left and right align within a scrolling parent like a modal.

screen shot 2017-11-09 at 11 05 01 pm
screen shot 2017-11-09 at 11 06 53 pm
screen shot 2017-11-09 at 11 21 32 pm
screen shot 2017-11-09 at 11 08 34 pm

PR #1145, opened 2017-10-24
Accordion component

### LINKED ISSUE #1033: Add SLDS Accordion to design-system-react
This is a placeholder issue to create a Salesforce Lightning Design System Accordion class as described at https://www.lightningdesignsystem.com/components/accordion/ to design-system-react.

This is so the community can keep track of what may be added. That way, the community can also offer suggestions or let me know if there is already some existing effort in progress.

lexishanson:
~~TODO: docs, snapshot tests.~~

@interactivellama I’m not sure I have a full understanding of snapshot test best practices. I’ll wait for your feedback on the snapshot test PR for Avatar ([#1143](https://github.com/salesforce/design-system-react/pull/1143)) and read through some other snapshot examples before writing these for Accordion.

Resolves #1033

### DIFF
--- a/components/accordion/index.jsx
+++ b/components/accordion/index.jsx
@@ -14,15 +14,22 @@ import { ACCORDION } from '../../utilities/constants';
 
 const propTypes = {
 	/**
-* Custom CSS classes added to `slds-accordion` node.
-*/
+	 * CSS class names to be added to the accordion container element. _Tested with snapshot testing._
+	 */
 	className: PropTypes.oneOfType([PropTypes.array, PropTypes.object, PropTypes.string]),
+	/**
+	 * HTML id for accordion component. _Tested with snapshot testing._
+	 */
 	id: PropTypes.string,
+	/**
+	 * The panel content for the Accordion component. Event handler for the accordion panels should be added here. Optional `panelContentActions` component may be passed as prop. _Tested with Mocha framework.
+	 */
 	children: PropTypes.node.isRequired
 };
+
 /**
  * An accordion allows a user to toggle the display of sections of content.
- * The accordion component wraps accordion items that can be selected and expanded.
+ * The accordion component wraps accordion panels that can be selected and expanded.
  */
 class Accordion extends Component {
 	componentWillMount () {
--- a/components/accordion/index.jsx
+++ b/components/accordion/index.jsx
@@ -1,0 +1,48 @@
+/* Copyright (c) 2015-present, salesforce.com, inc. All rights reserved */
+/* Licensed under BSD 3-Clause - see LICENSE.txt or git.io/sfdc-license */
+
+// Implements the [Accordion design
+// pattern](https://www.lightningdesignsystem.com/components/accordion/) in
+// React. Based on SLDS v2.4.3
+
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import shortid from 'shortid';
+
+import { ACCORDION } from '../../utilities/constants';
+
+const propTypes = {
+	/**
+* Custom CSS classes added to `slds-accordion` node.
+*/
+	className: PropTypes.oneOfType([PropTypes.array, PropTypes.object, PropTypes.string]),
+	id: PropTypes.string,
+	children: PropTypes.node.isRequired
+};
+/**
+ * An accordion allows a user to toggle the display of sections of content.
+ * The accordion component wraps accordion items that can be selected and expanded.
+ */
+class Accordion extends Component {
+	componentWillMount () {
+		this.generatedId = shortid.generate();
+	}
+	/**
+	 * Get the accordion's HTML id. Generate a new one if no id present.
+	 */
+	getId = () => this.props.id || this.generatedId;
+
+	render () {
+		return (
+			<ul name={this.getId()} className={classNames('slds-accordion', this.props.className)}>
+				{this.props.children}
+			</ul>
+		);
+	}
+}
+
+export default Accordion;
+
+Accordion.displayName = ACCORDION;
+Accordion.propTypes = propTypes;
--- a/components/accordion/index.jsx
+++ b/components/accordion/index.jsx
@@ -20,7 +20,7 @@ const propTypes = {
 	/**
 	 * HTML id for accordion component. _Tested with snapshot testing._
 	 */
-	id: PropTypes.string,
+	id: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
 	/**
 	 * The panel content for the Accordion component. Event handler for the accordion panels should be added here. Optional `panelContentActions` component may be passed as prop. _Tested with Mocha framework._
 	 */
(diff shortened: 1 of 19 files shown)


components/accordion/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+ * The accordion component wraps accordion panels that can be selected and expanded.
+ */
+class Accordion extends Component {
+	componentWillMount () {
+		this.generatedId = shortid.generate();
+	}
+	/**
+	 * Get the accordion's HTML id. Generate a new one if no id present.
+	 */
+	getId = () => this.props.id || this.generatedId;
```
Vintik: Why the method for this? Why not just inline it into name={}?

lexishanson: Good point, I will inline it since I'm only using it once.

interactivellama:
I'm not sure I fully understood how I had set up snapshot tests myself. ;-) 

Once #1150 is merged in, we should have deep DOM renders (instead of shallow) for components that use the `testDOMandHTML` method. I've removed Enzyme, since it's no longer needed once the overlay/dialog rewrite happens. Once #1140 is merged in we can do DOM-less tests without JSDOM (a DOM without a browser). 

DOM snapshots are helpful because you know what events and variables are being being passed down through "the web page" in memory. At their core, our components are templates with mouse/keyboard events attached to them, if something changes in the DOM snapshot, then we most likely changed the component. We will still need to do manual testing and automated event testing in the Mocha framework, but many times you want to re-factor something, you'll want to make sure nothing is changed (or only a little is changed) and snapshot testing is a fast way to know if we changed something. The HTML render is just a nice convenience to compare with the SLDS site.

As to what to test, there is always more to test. Most of the time though, you are wanting to test what the components "looks like" after a prop changes, so every prop changes requests a different test to verify that the correct thing changed.

components/accordion/private/panel.jsx
```diff
@@ -0,0 +1,78 @@
...
+	 * Summary header of the item belonging to this panel. _Tested with Mocha framework and snapshot testing._
+	 */
+	summary: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+	/**
+	 * HTML title attribute. _Tested with snapshot testing._
+	 */
+	title: PropTypes.string
+};
+
+const Panel = ({ children, expanded, id, panelContentActions, summary, title, onTogglePanel }) => (
```
interactivellama: This file shouldn't be private.

components/accordion/private/panel.jsx
```diff
@@ -0,0 +1,78 @@
...
+	/**
+	 * Callback that will run whenever a panel is toggled. Function should handle state to toggle `expanded` prop. _Tested with Mocha framework._
+	 */
+	onTogglePanel: PropTypes.func.isRequired,
+	/**
+	 * Summary header of the item belonging to this panel. _Tested with Mocha framework and snapshot testing._
+	 */
+	summary: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+	/**
+	 * HTML title attribute. _Tested with snapshot testing._
```
interactivellama: Please tell folks what element this is on--the summary? The purpose would be great, too. "The summary is truncated and this title allows the user to view the entire summary."

components/button/index.jsx
```diff
@@ -27,6 +27,10 @@ const Button = createReactClass({
 		 */
 		'aria-describedby': PropTypes.string,
 		/**
+		* Establishes a relationship between an interactive parent element and a child element to indicate which child element a parent element affects. Frequently used in cases where buttons or tabs are associated with exposing expandable regions.
```
interactivellama: Nice description!

examples/accordion/base.jsx
```diff
@@ -0,0 +1,144 @@
+import React from 'react';
+import Accordion from '../../components/accordion'; // `~` is replaced with design-system-react at runtime
```
interactivellama: Please use `~/` in examples for doc site compatibility

examples/accordion/base.jsx
```diff
@@ -0,0 +1,144 @@
...
+import Panel from '~/components/accordion/private/panel'; // `~` is replaced with design-system-react at runtime
+import IconSettings from '~/components/icon-settings'; // `~` is replaced with design-system-react at runtime
+import Dropdown from '../../components/menu-dropdown'; // `~` is replaced with design-system-react at runtime
```
interactivellama: Please use `~/` in examples for doc site compatibility

examples/accordion/base.jsx
```diff
@@ -0,0 +1,144 @@
...
+			<Dropdown
+				align="right"
+				id="ButtonGroupExampleDropdown"
+				assistiveText="More Options"
+				buttonVariant="icon"
+				buttonClassName="slds-shrink-none"
+				iconName="down"
+				iconVariant="border-filled"
+				onSelect={(option) => {
+					if (option.label === 'delete') {
```
interactivellama: This is pretty cool.

examples/accordion/base.jsx
```diff
@@ -0,0 +1,144 @@
...
+						value: 'C0'
+					}
+				]}
+				iconSize="x-small"
+			/>
+		);
+	}
+
+	togglePanel (event, data) {
+		this.setState((state) => ({
```
interactivellama: Really digging the `setState` callbacks...

tests/accordion/accordion.browser-test.jsx
```diff
@@ -0,0 +1,249 @@
...
+				...state.expandedPanels,
+				[id]: !state.expandedPanels[id]
+			}
+		}));
+	}
+
+	render () {
+		return (
+			<IconSettings iconPath="/assets/icons">
+				<Accordion id="base-example-accordion" openOn="click">
```
interactivellama: `openOn="click"` This seems out of place.

tests/accordion/accordion.browser-test.jsx
```diff
@@ -0,0 +1,249 @@
...
+
+AccordionExample.displayName = 'AccordionExampleComponent';
+AccordionExample.propTypes = propTypes;
+AccordionExample.defaultProps = defaultProps;
+
+/* Accordion rendering tests
+ */
+
+describe('Accordion', function () {
+	describe('Renders Accordion', () => {
```
interactivellama: These could easily be snapshots tests--that is "is X in the markup what it should be". Even the mouse actions, since we don't have state these could probably be snapshot tests-- that is we don't need to test the wrapping state container, just that when `isOpen`is true then the panel is open, etc.

interactivellama:
Storybook check complete.

Please align the details contents and summaries with https://www.lightningdesignsystem.com/components/accordion so the doc site example will look the same. If you'd like to keep the current one with lorem ipsum, feel free to have two Storybooks examples.

examples/accordion/base.jsx
lexishanson: Oof, sorry for the inconsistencies there. I also didn't know the reason we do this is for doc site compatibility. Good to learn.

components/button/index.jsx
donnieberg: +1 👍

tests/accordion/accordion.browser-test.jsx
donnieberg: yes, maybe this was just code left over from a copy/paste from another component? We should remove it in all the examples.

components/accordion/private/panel.jsx
lexishanson: Thanks for this! I tested and realized it wasn't truncating. I figured out the extra class it needed since it's using flexbox.

components/button/index.jsx
lexishanson: Thank you!

lexishanson:
Here are my updates to the Accordion component. Thank you both for the review @interactivellama @donnieberg.

@donnieberg I believe the nesting behavior should work, except there is issue. The `slds-is-open`class was updated by SLDS to only apply to the first child. The version of SLDS that DSR is on doesn't include this update yet, so the current `slds-is-open` class opens all of the nested accordion panels. I can build a nested accordion example as soon as the version is bumped.

PR #1150, opened 2017-10-27
Make Jest tests domless

interactivellama:
This paves the way for this library to work in a DOM-less environments--not that you'd want to UI components to work completely there, but this PR allows snapshot testing of applications if the consuming dev desires. There still may be components that are not compatible with a DOM-less environment, but ones with snapshot tests are.

Last four commits on top of #1140 

- Remove ButtonStateful extends TooltipTrigger: Delete tooltip/trigger (This was the only application using it)
- Guard DOM functions with conditions: This allows components to be rendered in an environment without a DOM.
- Make jest environment not use JSDOM: JSDOM environment has a document, etc. but React test renderer does provide DOM nodes for instances with React.findDOMNode() which is used by react-onclickoutside. react-onclickoutside tests for a DOM which JSDOM has, so it runs but fails.
- Switch to deep render for DOM snapshots

PR #1152, opened 2017-10-30
Add Combobox container class name prop

interactivellama:
Outer tag of Combobox doesn't have className prop. This adds it. This helpful for spacing, margins, etc. 

Last commit is the only commit of interest. The rest is built on #1150


examples/combobox/snapshot/base-open-class-name.jsx
```diff
@@ -0,0 +1,78 @@
...
+			selection: []
+		};
+	}
+
+	render () {
+		return (
+			<IconSettings iconPath="/assets/icons">
+				<Combobox
+					className="this-is-the-input"
+					classNameContainer="this-is-the-container"
```
donnieberg: hm, so I think in other components we use `containerClassName` (ie. app launcher, icon, selected-listbox). Can we change this and the other to `menuClassName`?

interactivellama: I'm worried about discoverability when they are alphabetical. What do you think about changing the other ones instead?

PR #1159, opened 2017-11-02
Fix story Lookup standard w/ accounts not loading

Vintik:
The lookup still looks a bit off, but from my understanding this is now deprecated? Either way thought we should still make sure the storybook for it at least loads.


interactivellama:
Thanks, Victor!

PR #1178, opened 2017-11-17
Make icons settings webpack friendly

kevinlambert:
When using webpack it's a pain to use the icons. You need to either manually copy the sprites or use a copy loader. 

Allowing the sprites to be assigned in the icon settings means webpack can re-write the urls and you don't need to co-ordinate a copy and path assignment.

```javascript
import utilitySprite from '@salesforce-ux/design-system/assets/icons/utility-sprite/svg/symbols.svg';
import customSprite from '@salesforce-ux/design-system/assets/icons/custom-sprite/svg/symbols.svg';

() => (
    <IconSettings utilitySprite={utilitySprite} customSprite={customSprite}>
    
    </IconSettings>
)
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+	
+[*]
+indent_style = tab
+trim_trailing_whitespace = false
+max_line_length = 140
+end_of_line = lf
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,6 +2,7 @@ root = true
 	
 [*]
 indent_style = tab
-trim_trailing_whitespace = false
+trim_trailing_whitespace = true
 max_line_length = 140
 end_of_line = lf
+insert_final_newline = true
--- a/components/icon-settings/index.jsx
+++ b/components/icon-settings/index.jsx
@@ -4,6 +4,25 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import { ICON_SETTINGS } from '../../utilities/constants';
+
+/**
+ * The Icon Settings component allows for the path to the icons to be specified.
+ * This should be used at the root of the application.
+ *
+ * **Individual sprites**
+ * If you are using webpack it is advised to use the sprite properties
+ * {actionSprite, standardSprite...} to specify the individual sprite paths so that webpack can
+ * easily re-write the paths.
+ * ```
+ * import actionSprite from '......';
+ *
+ * <IconSettings actionSprite={actionSprite} ......>
+ * ```
+ * **Root icon path**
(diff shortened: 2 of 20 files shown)


.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+	
+[*]
+indent_style = tab
+trim_trailing_whitespace = false
```
interactivellama: Probably needs 
```
trim_trailing_whitespace = true
insert_final_newline = true
``` 
if you meant to PR this file.

components/icon-settings/index.jsx
```diff
@@ -6,7 +6,14 @@ import PropTypes from 'prop-types';
 
 class IconSettings extends React.Component {
 	getChildContext () {
-		return { iconPath: this.props.iconPath };
+		return {
+			iconPath: this.props.iconPath,
+			actionSprite: this.props.actionSprite,
```
interactivellama: I like this. Kind of aligns with the [anti-concatenation approach](https://github.com/salesforce/design-system-react/blob/master/docs/codebase-overview.md#all-text-can-be-internationalized) of this library.

interactivellama: * Do you think there's a way to make it more obvious that it's the path + file name. The `propType` is a string, so that should help. 
* Also could you add an example/storybook story that uses this in the `examples/icon` folder?

* I know we failed to have already didn't have a propTypes object on `IconSettings`. Could you add the props you added to a `IconSettings.propTypes` object add comment descriptions for them. Please look at other components for examples.

I think this will be helpful for devs, so they can put their icons anyway they want.

interactivellama: I'll pull this branch down on Monday and make sure it works with the story I've requested. Thank you for your contribution. I'll try to get a release out for you next week. It will mostly likely be 0.8.0 that uses the new positioning library, PopperJS.

.editorconfig
kevinlambert: Yes I meant it to be in the PR. Added the changes.

components/icon-settings/index.jsx
kevinlambert: Thanks @interactivellama for the comments

I've added:
- an Icon Settings example.
- the new icons setting option to the readme.
- the prop types to icon settings.

Note that I've added a new constant `SLDSIconSettings` so this is what is rendered as the element tag.

Jest Snapshots have been updated.

With regards to the sprite properties on the icon settings, it could be named `actionSpritePath` but that may be even more confusing cause in reality you are passing an svg object and webpack is extracting that out and replacing it with the path string. I think with an icon settings example and the documentation it should be clear enough to devs what to do.

Let me know if there are any other changes that need to be done.

Issue #877, opened 2017-02-13
Clicks on page with datepicker go to top of page


interactivellama:
Datepicker click issue. Clicking goes to the top of the page right now. URL hashes do not work either. https://react.lightningdesignsystem.com/components/datepickers/

This happens to any page with a datepicker on it.

PR #1187, opened 2017-11-24
Focus input if menu was actually open and not just requested to close

keklesurvivant:
Hello,

I faced an issue with Datepicker which was already referenced in an old ticket (#877).
I made the following change : for me the focus to the input should be done only if the related Dialog was previously opened to avoid unexpected effect if the Datepicker was not being used.
Sorry, I'm not really accustomed to contributing to opensource projects in github. Do not hesitate if I miss a step in the contributing process

Related to #877

Kévin

### DIFF
--- a/components/date-picker/date-picker.jsx
+++ b/components/date-picker/date-picker.jsx
@@ -313,10 +313,10 @@ class Datepicker extends React.Component {
 
 		if (this.getIsOpen()) {
 			this.setState({ isOpen: false });
-		}
-
-		if (this.inputRef) {
-			this.inputRef.focus();
+			
+			if (this.inputRef) {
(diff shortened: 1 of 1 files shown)


interactivellama:
I'll look into code next week. You might consider something like [componentDidUpdate](https://reactjs.org/docs/react-component.html#componentdidupdate) and test if the menu was open previously and is now is closed based on a compare of `this.state` and prevState/prevProp. Combobox has some examples of `componentDidUpdate` if you want to look at it.

As you know, Datepicker can be controlled with props or uncontrolled with state as to whether the menu is open, so that makes it a little trickier.

interactivellama:
@keklesurvivant Moving to #1341

Issue #1146, opened 2017-10-25
Publish Design System React to NPM


vernak2539:
Can we get this lib published to NPM? It's kind of a headache to link to it via git url

interactivellama:
The first purpose of open sourcing was to eliminate the access issue for folks. NPM publishing is in the plans. I don't have an ETA right now. I am evaluating the file structure of the module. Comments are welcome.

## NPM module file structure

**esnext/jsnext:main/module/*.mjs (“Michael Jackson Script”)...**

Background

* https://nodesource.com/blog/es-modules-and-node-js-hard-choices/#currentstateofdiscussion (from 2016)
* Can we make the main dsr release (`@salesforce/design-system-react`) have both the dist files and the src files, with the src files referenced by `jsnext:main`? 
* Could we potentially release "cleaner" versions that are postfixed? (`@salesforce/design-system-react-cjs`, `@salesforce/design-system-react-ejs`, `@salesforce/design-system-react-amd`, etc)
* Could we do subfolders? - `@salesforce/design-system-react` main points to `@salesforce/design-system-react/cjs` but you can also access `@salesforce/design-system-react/ejs`
* CommonJS transpile in root folder (default for node modules, linked to by package.main), require is synchronous
* ES6 modules transpiled in es folder (No JSX, no ES6 classes), linked to by package.jsnext:main and package.module), import is asynchronous and allows tree shaking by module builder

In a perfect “breaking changes” world

* dist/design-system-react.js (bundle)
* es/ (transpiled, except for import/export to allow tree shaking, no babel presets needed)
* src/ complete production source code
* lib/ commonjs transpiled (no babel presets needed)

A possibly non-breaking change proposal

* [root]/design-system-react.js (bundle, backwards compatible)
* [root] (ES6 tag file structure, transpiled, except for import/export to allow tree shaking, no babel presets needed, backwards compatible)
* src/ complete production source code (has JSX, examples folder, etc.)
* lib/ commonjs transpiled (no babel presets needed, linked to from package.main, possibly backwards compatible)

Benefits of non-breaking change proposal

* Doesn't break imports used in interactive examples for two years.
* Should be foreseeable future proof
* Allows CommonJS via single entry point 
    * Is CJS needed? Who builds React apps without a transpiler and a module builder?
* Ignore AMD version (still published from repo tag).
* UMD version is the bundled file.

interactivellama:
Consider new or no node engine spec for npm package. #1175

[...]

PR #1195, opened 2017-12-08
DO NOT MERGE: Adds a command for files in an npm package

futuremint:
Fixes issue #1146 

### Additional description

Please review the output of running "npm run dist" in this PR.  It adds an additional temp directory called `.tmp-npm`.  I intend to publish the npm package from that folder.  In its current state this PR uses Babel to down-compile what is in `.tmp-es` to general ES6.  The idea is that we don't want to force any particular babel presets onto our users, but that we can be fairly sure they're using _some kind of_ JS compiler in their toolchain downstream from this package to target different JS versions (such as ES5 or whatever they want).

---
### Reviewer checklist
- [ ] While the contributor's branch is checked out, run `npm run dist` and compare the generated JS files to the source JSX files.

### DIFF
--- a/scripts/dist.sh
+++ b/scripts/dist.sh
@@ -41,6 +41,7 @@ echo "## Copying the components"
 cp -r .tmp .tmp-amd
 cp -r .tmp .tmp-commonjs
 cp -r .tmp .tmp-es
+cp -r .tmp .tmp-npm
 rm -rf .tmp/
 
 cp -r components .tmp-es/components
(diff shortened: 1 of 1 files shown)


interactivellama:
I think this is looking good. We can always go _more_ transpiled. We can't go less though in the future. The "@salesforce" namespace needs to be added to the package, so devs don't have to change that in the future. I'm going to test with create-react-app today.

interactivellama:
The update on this is that we are hoping to do multiple versions in one tag:
* a lightly transpiled ES6 that maintains modules in `/components/`
* a CJS version that means if you use `import { Button } from 'design-system-react'`, you won't need Babel transpiling of files in `node_modules` which will make this format compatible with a non-ejected `create-react-app`. __CJS can be swapped out for "modular ES5" in this instance, I think.__

interactivellama:
Closing in favor of #1210

Issue #945, opened 2017-03-23
Look into AST parser that formats code at git staging


interactivellama:
Get on the prettier train! https://github.com/prettier/prettier
from ReactConf '17 http://conf.reactjs.org/speakers/james

![screen shot 2017-03-23 at 3 14 17 pm](https://cloud.githubusercontent.com/assets/1290832/24270595/690fbda0-0fdb-11e7-9e41-de9c7f89e118.png)

The question is do we want this on commit or on staging/add? https://github.com/prettier/prettier#pre-commit-hook-for-changed-files

We probably could hijack the staging commands of our IDEs as the shortcut for this parsing task.

interactivellama:
Closing for now. I really enjoy the experience of linting at commit and may see if I can get something optional working in the future.

PR #1196, opened 2017-12-08
Add Prettier and eslint fix support

interactivellama:
Fixes issue #945

### Additional description
This does not do anything at git-staging. I'm curious to when devs would want their code "re-written." The simplist way is `npm run format` that has been added which will run prettier and _then also_ run `eslint --fix`, but ESlint plugin and format on save with Prettier plugin will work in VSCode for instance.

### Tasks completed
* Audited ESlint settings from old eslint-contrib-slds repo that no one uses to align with DSR code and not Marketing Cloud's UI platform team's codebase which is what it was before.
* When in doubt, let AirBnB decide with presets
* I found a bug in `PageHeader` that used `RelatedList` as a render component, but `RelatedList` was never imported, so it was undefined ALWAYS. This made me want to remove undefined and update babel-eslint and eslint to know about class properties. 🎉 
* Updating libraries also allowed glob file overrides for tests and examples, so 60+ `.eslintrc` could be removed that were added in my recent "Group Like Files" PR. 🎉 

The PR is huge because it also has the changes to all the js/jsx files in it, too.

### Things I learned from this task
* Prettier is for styling. The awesome part is that it abstracts to AST and then brings it back in source code. Prettier is FB's opinionated code style engine, it's opinions can change on minor upgrades, not just major, so it's fixed to patches at 1.9.x.
* ESlint should be used for code quality issues, limiting code to "the good parts," check paths and dependencies, etc. It can be helpful for putting dependencies in their correct spot. I found one lodash library that is not listed in any of the dependencies and must be available, because of another module, since it didn't crash anything!
* I've grown dependent on allowing ESlint to keep track of some types of things (wrong import path for example). Loose eslint rules + trusting the linter is not good. These tools have grown powerful recently. I respect them for more than "comma style wars", now that I know them better.
* eslint and Prettier can conflict. One example is use of + and - operators. ESlint/AirBnB presets say you always need parens for `+` and `-` operators. Prettier only wants them if `||` and `&&` or `+` and `*` are used and the operators aren't on the same level for order of operations.
* Our ESlint settings and Prettier "space before function parameter parens" conflict. There may be some others, but that's the big one. 
* I'm open to comma dangle/ trailing comma and it just takes one command to switch the whole codebase now. I haven't switched it though.
* I've left the JSON files to not be formatted. Prettier does try though. I'd be open to formatting them though.


interactivellama:
This PR is currently for transparency. I will look into smashing these into a config commit and a code lint commit for easier review.

interactivellama:
This is ready for review @cormacmccarthy and no, I'm not just making 500+ file PRs to make the line addition/substractions more than @donnieberg and @ivanbogdanov 😄 

This change should allow `npm run format` to be run before someone pull requests or use IDE extensions to format code on save.

interactivellama:
We will wait until #1199 and #1200 are merged to merge this one to avoid merge conflicts. Maybe sometime around Christmas. Should be just one command to update.

Issue #1206, opened 2017-12-16
implementing dynamic menu dsr


woshichensida:
Hi, per contributor guide, I am creating an issue to implement dynamic menu dsr. The component slds is [here](https://latest-212.lightningdesignsystem.com/components/dynamic-menu/#content).
As part of our work on previous release, we already implement the over 10 items variation with search box in our repo iot-ui. And we are planning to move the code over and start from there. Let me know what you guys think. Thanks

interactivellama:
I'll try to get you what I'd think the API would like as a DSR component on Monday and we can compare notes. Thanks so much! Looking forward to it.

interactivellama:
This is what I was thinking after 15 minutes:

```
const propTypes = {
    assistiveText: shape({
        popoverLabel: PropTypes.string
    }),
    children: PropTypes.node, // trigger/Button
    classNamePopover: PropTypes.oneOfType([PropTypes.array, PropTypes.object, PropTypes.string]),
    combobox: PropTypes.node, // Combobox component
    hasStaticAlignment: PropTypes.bool,
    isOpen: PropTypes.bool,
    menuItem: PropTypes.func, // SLDS list item shown should be default, but this custom render prop is optional to include
    popoverMenuItem: PropTypes.func, // SLDS list item shown should be default, but this custom render prop is optional to include
    popoverPosition: PropTypes.oneOf(['absolute', 'overflowBoundaryElement', 'relative']),
    popoverOptions: PropTypes.array
};

const Example = () => (
    <DynamicMenu
        combobox={
            <Combobox
                id="combobox-unique-id"
                labels={{
                    placeholder: 'Search Favorite',
                    noOptionsFound: 'You can favorite any page!',
                    noOptionsFoundHeading: 'My Favorites'
                }}
                options={accountsWithIcon} // Combobox will probably need a heading list item added to it, that is removed from the a11y keyboard interactions and events
            />
        }
        popoverOptions={[
            <Button label="Favorite this page" onClick={function () {}} icon={<Icon category="utility" name="add" />} />,
            <Button label="Favorite this page" onClick={function () {}} icon={<Icon category="utility" name="edit" />} />
        ]}
    >
        <Button icon={<Icon category="utility" name="favorite" variant="icon" />} />
    </DynamicMenu>
);
```

[...]

PR #1208, opened 2017-12-27
Dynamic Menu

lexishanson:
WIP - pull request here to review design. 

Lots of todos left, but should provide general idea of direction we're heading.


lexishanson:
Hi @interactivellama here are some details and thoughts on what I did in the last few commits:

**Menu**
Added `isDynamicMenu` prop. This prop is responsible for applying certain classes in the case of a DynamicMenu. An alternative option to creating this prop is to pass in custom classnames to Menu component, but this seems more complicated - when it  just a regular Menu, some classes should apply that shouldn’t apply when it is a DynamicMenu.

**Popover**
Added logic, for when Popover is open, to conditionally render “close” button. This seems like it could be useful outside of our use case. I read in the docs site that the close button is required in `Popover`, but this wasn’t listed as a requirement in SLDS. Maybe this changed recently?

Added `bodyClassName` prop to allow for custom classes to be passed  to the body in popover. This could also be useful outside of DynamicMenu use case. Specifically for DynamicMenu example, it needs ”slds-p-horizontal_none" as a custom classname to remove padding.

**Button**
Added `IconVariant: […, ”dynamic-menu”]`. This handles adding specific classes to the buttons on the footer of the popover for DynamicMenu. Alternative here is for the dev to pass in custom classnames.

**Combobox**
Added prop for `hideInput` to Combobox. 



----------------------




While I think most of the additions above are reasonable, the case is getting stronger for a separate `DynamicMenu` component. 

For context, here are the DynamicMenu Variants:

- 	Base
- 	Items (base)
- 	1 Item
- 	10 or Fewer Items (no Combobox, no scroll)
- 	More Than 10 Items (Combobox, scroll)

For the variants other than the “More than 10 items” variant we’ve been building for, we have a couple of other flags to think about:

**Popover**
A prop would need to be added to handle rendering the popover header. Per SLDS, the header should only render when Combobox does not render (10 items or less). The other option is to give control to the dev on this and show it in examples.

**Combobox**
According to SLDS, the Combobox should disappear when there are fewer than 10 items. In the API we’re following currently, we should either: 1) hide just the input and not the Combobox container, since the listbox and menu are inside of it (though markup wouldn’t match SLDS) or 2) render just the listbox and menu on `renderDynamicMenu` in Combobox. Option 2 is a little weird, since we’re not outputting a Combobox from the Combobox component.

What are your thoughts based on the other variants we'll want to build at some point? Do you think we can still make `DynamicMenu` a variant of `Combobox`?

interactivellama:
I think DynamicMenu as it's only component provides clarity. The API should be pretty close to what it is now. The underlying infrastructure you and Sida have created (variants, classnames, etc.) are well thought out and most likely just need some packaging up.

Do you have any feedback or issues with the [second API proposal](https://github.com/salesforce/design-system-react/issues/1206#issuecomment-352319226)?

lexishanson:
@interactivellama Thanks for the feedback. For the second API proposal, are you referring to: 
```
<DynamicMenu>
  <Popover>
    <Combobox
      variant="dynamic-menu"
      actions={[<Button label="Add favorite" />]}
    />
  </Popover>
</DynamicMenu>
```

What do you think about us keeping the change to add the `bodyClassName` prop to `Popover` and to allow for the option to remove the "close" button in `Popover`? Should we just handle this within `DynamicMenu`?

interactivellama:
If a UX pattern is an "always" (such as remove the popover close button and use the special `bodyClassName` then it should be built into DynamicMenu. This will reduce/simplify "always on / always the same" APIs from the consumer's source code.

interactivellama:
I'm going to close this until it can be picked back up again.

Issue #1213, opened 2018-01-11
Lookup throw error on input focus


NuclleaR:
Dialog component needs position prop and marked as required but in Lookup component this prop is missing. So Dialog doesn't render.

interactivellama:
Fixed by #1214.

@NuclleaR https://github.com/salesforce/design-system-react/releases/tag/v0.8.4 is out with the fix. We now have NPM modules, too https://www.npmjs.com/package/@salesforce/design-system-react

PR #1214, opened 2018-01-11
Add `dialogPosition` prop to Lookup as required for Dialog

NuclleaR:
Fixes issue #1213

### DIFF
--- a/components/lookup/lookup.jsx
+++ b/components/lookup/lookup.jsx
@@ -601,6 +601,10 @@ const Lookup = createReactClass({
 	},
 
 	renderSeparateMenu () {
+		// FOR BACKWARDS COMPATIBILITY
+		const menuPosition = this.props.isInline
+			? 'relative'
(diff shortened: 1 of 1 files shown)


interactivellama:
Please update to `menuPosition` to align with [other components](https://github.com/salesforce/design-system-react/search?utf8=%E2%9C%93&q=menuPosition&type=)

NuclleaR:
@interactivellama Could you pls review again

interactivellama:
One more thing, please add something like:

```
// FOR BACKWARDS COMPATIBILITY
const menuPosition = this.props.isInline ? 'relative' : this.props.menuPosition; // eslint-disable-line react/prop-types
```
right before `<Dialog/>` to make it backwards compatible for folks. Here's an example https://github.com/salesforce/design-system-react/blob/master/components/combobox/combobox.jsx#L380

interactivellama:
@NuclleaR Thank you for the fix!

PR #1223, opened 2018-01-18
Improve dev experience

cormacmccarthy:


interactivellama:
* When are folks supposed to run Prettier? I don't see it run in the CI.
* Is it better to have 0 warnings and have to search for `eslint-disable` in order fix things?

`npm run format` is the primary one you want to run. We could make it `lint:fix` though.

package.json
```diff
@@ -28,6 +28,7 @@
     "link": "./scripts/npm-link.sh",
     "lint": "./node_modules/.bin/eslint \"**/*.{js,jsx}\"",
     "lint-fix": "./node_modules/.bin/eslint --fix \"**/*.{js,jsx}\"",
+    "lint:fix": "./node_modules/.bin/eslint --fix \"**/*.{js,jsx}\"",
```
futuremint: While I agree with this new name format, it doesn't match the format of the rest of the script names.  Put this back to `lint-fix`, or change the names of _all_ the scripts to have `:` instead of `-`s.

package.json
```diff
@@ -203,12 +203,6 @@
     "webpack-hot-middleware": "^2.18.2",
     "xml2json": "^0.11.0"
   },
-  "pre-commit": [
-    "lint-staged"
```
futuremint: I agree with this change... linting should be checked by the CI after push, but _before_ merging.

package.json
```diff
@@ -203,12 +203,6 @@
...
-  ],
-  "pre-push": [
-    "validate"
```
futuremint: I agree with this change... CI should do this after push, but _before_ merging. This will also make invalid code much more visible, since it'll be a badge at the bottom of the PR on the PR screen for everyone to see.

futuremint:
I am sad to see all of the a11y disabled lines. However I will say that I think this is the right approach... change the settings to that we start in a clean state now (0 errors, 0 warnings), and we need to make "lint cleanup issues" to capture some of the problems (i.e. all of the ignored a11y settings) and work on them piece by piece when prioritized for sprint work.

cormacmccarthy:
> I am sad to see all of the a11y disabled lines.

Yep! I totally agree. I want to take a moment to point out that I didn't global disable the a11y stuff. I went out of my way to go file by file and disable it on a line by line basis. This way we can't add _new_ errors, but we need to be realistic about our ability to fix the errors we have right now. Sometimes cleaning your room means just boxing everything up and stacking the boxes in the closet until you can take them out and deal with them one at a time.

Seeing 800+ warnings in my terminal is like hearing there are 800+ snakes crawling all over my code. I can't fix that let alone function optimally with that knowledge being shoved in my face. Let me put those snakes in boxes until I can deal with them instead of having to have them crawl all over my lap while I try and work.

As a good faith show at my willingness to eventually get these boxes out and deal with the snakes, I present: #1224

cormacmccarthy:
> Is it better to have 0 warnings and have to search for eslint-disable in order fix things?

It is better to have 0 warnings. But at the point where you want to fix things, you should flip the toggle in the eslint settings and see all the warnings and start fixing them.

I'd happily approve a PR where someone uses some sort of tool that takes all failing files and puts line-specific `// eslint-disable-line` comments into them, or `/* eslint-disable */` comments at the top.

The goal here is that we should never ever ever have warnings or errors that we are ignoring (not just in eslint either! Across the entire project, when we run automation tasks there should be no warnings or errors spitting out to the console). We can argue all day about the difference between an error and a warning (eg: error being something so egregious that we can never accept a solution where new code overrides the error, and warning being something it is acceptable to override under certain circumstances) but the point is we shouldn't just be ignoring either of them. As soon as you make a habit of ignoring warnings and errors you have set yourself on the path to _accidentally_ ignore a warning or error that really can not be ignored, and then the code makes it through and goes live and we have a bug.

I've worked at jobs where a process would run, 100s of errors and warnings would spit out, and I'd go, "hey! wait! All that looks really bad!" and the response was, "oh, no it's probably nothing. You'll get used to the shape of it as it flies by. As long as the shape looks the same each time you're probably fine. It's when you have _new_ errors or warnings that you know there might be a problem". That's no way to live.

interactivellama:
Don't fee too bad. I'm not saying it's all of them, but the majority of a11y issues I have seen are a good rule that SLDS has chosen to override in a specific example and should be ignored/disabled.

components/page-header/private/title.jsx
```diff
@@ -1,6 +1,8 @@
 /* Copyright (c) 2015-present, salesforce.com, inc. All rights reserved */
 /* Licensed under BSD 3-Clause - see LICENSE.txt or git.io/sfdc-license */
 
+/* eslint-disable */
```
cormacmccarthy: Was this an accident??

interactivellama:
@cormacmccarthy If we remove `prettier-eslint` from conveniently informing lint issues at pre-commit, we'll need to add it to the CI to warn at that location. Right now we only fail on ESlint errors in `test.sh`.

Please add additional stricter tests to the CI server if you want to remove these from pre-commit.

futuremint:
Along those same lines: the CI server should be running a very obvious script that is exclusive to the CI environment (but composed of the usual scripts, such as test.sh, etc.).  This would make it easier to quickly understand what is going on for newcomers to the project.  I would propose a `travis-ci` script in `package.json` that points to a shell script (this would be for find-ability as much as anything else), and the shell script should be exclusive to CI, named `travis-ci.sh` or some such thing.  And that would contain no special logic, but just the list of commands that Travis runs.

Kind of like leaving a trail of breadcrumbs...

cormacmccarthy:
Done. Let me know specifically what you are hoping to see in there that isn't there currently.

PR #1224, opened 2018-01-19
Run react codemod transform class on files

cormacmccarthy:
So that we can enable the eslint `react/sort-comp` check.

This PR builds on (and includes) #1223 


interactivellama:
@cormacmccarthy Is this ready to be checked?

components/data-table/column.jsx
```diff
@@ -27,15 +27,15 @@ DataTableColumn.propTypes = {
 	 * import DataTableCell from 'design-system-react/data-table/cell';
 	 * const CustomDataTableCell = ({ children, ...props }) => (
 	 *   <DataTableCell {...props} >
-	 *     <a href="javascript:void(0);">{children}</a>
+	 *	 <a href="javascript:void(0);">{children}</a>
```
interactivellama: The tabs in the comments should be reverted or they will end up as tabs in the code examples in the doc site.

cormacmccarthy:
@interactivellama yes

.eslintrc.json
```diff
@@ -163,6 +163,19 @@
...
+    // Taken from eslint-config-sfmc
+    "func-names": 0,
+    "import/first": 0,
+    "react/no-multi-comp": 0,
+
+
+    // When someone is ready to tackle these issues, feel free to uncomment the following
+    // Do not push a change that causes lint errors/warnings to start appearing
+    // All lint warnings errors should be meaningful, and therefore, there should never intentionally be any
+    "react/prefer-es6-class": 0,
```
interactivellama: Did you just tackle this? sort comp too?

cormacmccarthy: No, these three settings can not be turned on without a substantial amount of re-work.

package.json
```diff
@@ -27,13 +27,13 @@
     "icons": "./node_modules/.bin/babel-node scripts/inline-icons.js",
     "link": "./scripts/npm-link.sh",
     "lint": "./node_modules/.bin/eslint \"**/*.{js,jsx}\"",
-    "lint-fix": "./node_modules/.bin/eslint --fix \"**/*.{js,jsx}\"",
+    "lint:fix": "npm run format &&./node_modules/.bin/eslint --fix \"**/*.{js,jsx}\"",
```
interactivellama: I'm not sure why these are all being changed. Are dashes a bad convention? 

I don't mind changing them, but you will need to update all the locations in the documentation that use them. Please do a search for all the commands here in the repo, so we don't have out-dated documentation. You will probably want to go through the doc site in the UX org to make sure that documentation is updated there, too.

cormacmccarthy: It isn't that they are bad convention so much as there is currently _no_ convention.

Right now, there is no difference between dash-as-space and dash-as-signal-that-we-are-passing-flags

So, does `snapshot-test` mean `snapshot --test` or does it mean `snapshot-test`? Right now there is no way to tell. With this change it becomes immediately obvious that `npm lint:fix` calls the `npm lint` task with a `--fix` flag (once you know the convention). 

Another example: before `publish-to-git` and `publish-to-edge` both called the same publish command, but that was not obvious unless you looked. Now, `publish` does the default publish, and `publish:edge` does publish but passes in a flag indicating that you are publishing to edge.

This also leaves room for a script to have both dashes-as-spaces _and_ still pass in flags and not mix the two together:

eg: the imaginary `snapshot-test:dom` would call the snapshot-test command, but pass in a flag indicating that we only wanted to test the dom. It would be obvious immediately that this was calling the same command as just `snapshot-test` but that we are passing in a flag. There would be no confusion as to whether this was a `snapshot` command that we were passing a `test` flag to (as there is now).

It's more self-documenting.

interactivellama: Good idea. I like the clarity. This PR is good to go. It's waiting on #1223 to complete, but the class transforms and eslint changes are good.

Issue #1099, opened 2017-09-27
Should have visual regression testing


interactivellama:
All changes should be visual regression tested to ensure quality. It should be automated, and should run as part of the current PR testing process.

The framework was added in #1241. Examples just need to be added and tested. This should probably be broken up into multiple issues and most used components prioritized.

interactivellama:
This should probably be broken up.

interactivellama:
We should start with these when we work on this issue:
* [] Combobox / Lookups / Picklist
* [] Data Tables
* [] Icons
* [] Input
* [] Menus
* [] Page Headers
* [] Tooltips
* [] Trees

futuremint:
Theoretically: if we have the ability to automatically validate that our markup matches what is specified in SLDS, do we need visual regression testing?  I ask because visual regression testing is slow, fragile, and complicated... and to keep a high quality developer experience in my opinion we should avoid it unless we absolutely need it.

interactivellama:
The most recent backstory on this is with that stories were manually checked with the CSS `--` to `_` BEM modifier change and `VerticalNav` broke and VRT would have blocked me from releasing that broke.

One option (which is similar to what SLDS does and is close to how a 3rd party VRT would work) is that the code review person would generate the snapshot image before code commits and after code commits and do a compare to better enable their code review. 

Large codebase changes and side effects (one component changes another) are the primary issues we are trying to highlight here.

interactivellama:
Closing with #1987 merged in and #2098 highlighting which components are missing from tests

PR #1241, opened 2018-01-25
Add automated image and DOM snapshots

### LINKED ISSUE #1050: Implement Storyshots
It's Jest DOM snapshots of all Storybook stories. https://github.com/storybooks/storybook/tree/master/addons/storyshots

Will implement after #951. Jest does't like TetherJS.

interactivellama:
Fixes issue #1099 and fixes #1050.

### Additional description
* This allows a new story in Storybook to automatically generate a DOM and an image snapshot.
* Adds an alpha version of https://github.com/storybooks/storybook/tree/master/addons/storyshots which has image snapshots (puppeteer/headless chromium) built into it.
* ~We still need concurrent servers running, so that has to be fixed. `npm run build-storybook && node app.js && npm run snapshot-test` should do it.~
* ~11 Storybook stories have been commented out due to errors (or not guarding variables) in our components. With the removal of TetherJS, we mostly have DOM (`document not found`) issues with our UI library that Jest doesn't understand. My proposal right now is to have two Storybook configs (or have an Node ENV variable that swaps the config). This would require the end dev to add an import of his/her stories to two configs, but not have to write image or DOM tests.~
* SLDS CSS can't be read by Jest right now, so removing the import from Webpack and linking to the CSS in the header on a page on an Express server seems like the only option.

**Update: All but three stories have been commented out for image snapshot testing. I believe it would be best to get the stories and components fixed/corrected before we commit MBs of images into this repository that will then change again in the future.**

---


.storybook-image-snapshots/config.js
```diff
@@ -0,0 +1,7 @@
+import { configure } from '@storybook/react';
+
+function loadStories() {
+	require('../components/storybook-stories-image-snapshots'); // eslint-disable-line global-require
```
cormacmccarthy: It's not obvious to me why you are doing this this way. Especially since you have to disable an eslint rule...

components/storybook-stories-image-snapshots.js
```diff
@@ -0,0 +1,48 @@
+// Sample stories for image snapshot testing
+// Components commented twice are not currently compatible with jest-image-snapshot
```
cormacmccarthy: Why not?

tests/initial-blank-stories.jsx
```diff
@@ -0,0 +1,14 @@
+/*
+ * This file exists to place a buffer between
+ * the Express server loading and image snapshots
+ * running.
```
cormacmccarthy: Why?

.storybook-image-snapshots/config.js
interactivellama: It's a CJS way of bringing in the variables. I'm open to another way. It was not a problem I was trying to solve.

The library is ES6, except when it's not--which is mostly in Node which is CJS. Until node supports imports. I figure we'll need it like this and so we have a eslint comment

components/storybook-stories-image-snapshots.js
interactivellama: Various reasons, but I mostly spelled this out in the Pull Request description. Many of our components measure things or create elements on the DOM.

tests/initial-blank-stories.jsx
interactivellama: I'm thinking we should maybe change this to a "Welcome to DSR" heading centered page which would be kind of nice as the first story. What do you think?

cormacmccarthy:
Master right now:
image

Your changes:
image

interactivellama:
Updated. Thanks.
screen shot 2018-01-28 at 10 49 41 pm

interactivellama:
@cormacmccarthy Updated. I also revised the `contributing.md` process to account for Storyshots.

PR #1243, opened 2018-01-25
Replace airbnb shape props with PropTypes.shape

kevinlambert:
Airbnb prop types is redundant with PropTypes.shape.

### Additional description

---


interactivellama:
My understanding is that `PropTypes.shape` requires all keys to be present, while Airbnb's allows for a subset to be present and only check to see that the keys are limited to the names specified. Is this false?

interactivellama:
I could be wrong but I think the React PropType one forces all the keys while AirBnB's only forces it on `non-null/undefined value`.

`shape: takes a shape, and allows it to be enforced on any non-null/undefined value.` 
See https://www.npmjs.com/package/airbnb-prop-types

interactivellama:
If it is different, sounds like a comment is needed in the source code for clarity. I'm going to close this and add a comment about non-null/undefined values unless we can verify that it's the same.

interactivellama:
@kevinlambert I looked into this a little further and it appears that the behavior of `PropType.shape` has changed to match the behavior of air-bnb's. I'm not the only one confused it looks like: https://github.com/facebook/react/issues/5916

Apologies for jumping the gun on this "not working" without actually running the code. I was having flashbacks to trying to get `shape` working for days back in 2016, I think.

I'll update the merge conflicts and get back to you. Thanks!

interactivellama:
Fixes: #1281

kevinlambert:
Cool. No problem. I was also a bit confused by the docs too. :)

Issue #1246, opened 2018-01-26
Update SLDS peer dependency to allow  >= 2.4.x


davidlygagnon:
Update SLDS peer dependency to allow  >= 2.4.x

PR #1247, opened 2018-01-26
Update SLDS peer dependency

davidlygagnon:
Fixes issue #

https://github.com/salesforce/design-system-react/issues/1246


interactivellama:
This is fine. We can bump it, once DSR starts wanting stuff in 2.5 or once the visual regression tests go live.

Issue #1252, opened 2018-02-02
GlobalHeader throw error when click Profile


NuclleaR:
When click on Profile nothing happens
in dev console
```
TypeError: Cannot read property 'props' of null
```

PR #1253, opened 2018-02-02
Fix GlobalHeader Profile MenuDropdown issue

NuclleaR:
Fix GlobalHeader Profile MenuDropdown issue

Fixes issue #1252


interactivellama:
Thanks!

PR #1261, opened 2018-02-05
Implement no plus plus eslint rule

interactivellama:
Increases code consistency. Removes eslint rule from config.

This one may require manual UI testing of affected components, due to possible refactor side effects. I have manually checked Lookup up/down key, App launch tile truncate, datepicker calendar, low priority warning works, progress indicator manually and they are working correctly.


components/date-picker/private/calendar.jsx
```diff
@@ -220,7 +220,8 @@ const DatepickerCalendar = createReactClass({
 
 			// create new weeks
 			firstDayOfWeek = DateUtil.addWeeks(firstDayOfWeek, 1);
-			done = count++ > 2 && monthIndex !== firstDayOfWeek.getMonth();
+			done = count > 2 && monthIndex !== firstDayOfWeek.getMonth();
+			count += 1;
```
futuremint: This is a great example of why to try to avoid the postfix operator. In the old code here, `>` has a higher operator precedence than `++` but the code is written in a way that it _looks like_ `count` is incremented _before_ checking if its greater than `2`. However `>` get compared _first_, so `count` gets "greater than 2" _first_, and _then_ gets incremented _after_ the comparison which makes the old version of this code potentially confusing.

components/progress-indicator/index.jsx
```diff
@@ -123,11 +123,16 @@ const defaultProps = {
...
 function checkSteps (steps) {
-	if (steps === undefined) return false;
-	for (let i = 0; i < steps.length; ++i) {
-		if (steps[i].label === undefined) return false;
+	const anyLabelUndefined = (step) => step.label === undefined;
+	let valid = true;
+
+	if (steps === undefined) {
+		valid = false;
+	} else if (Array.isArray(steps) && steps.some(anyLabelUndefined)) {
```
futuremint: If you move this expression and the one above to a local variable and invert the boolean test, then the entire `if/else if` block could be removed, the code could be more readable, and line 135 could be simplified. For example:
```
const labelDefined = (step) => step.label !== undefined;
const stepsDefined = steps !== undefined;
const stepLabelsDefined = Array.isArray(steps) && steps.all(labelDefined);

return stepsDefined && stepLabelsDefined
```

Meaning... you're returning `true` only if _both_ `steps` is defined _and all_ step labels are defined.

interactivellama:
Much clearer:
screen shot 2018-02-26 at 4 01 39 pm

Issue #1266, opened 2018-02-06
Popover is not trapping focus


interactivellama:
Popover should be moving focus and trapping it from tabbing out of Popover.

http://design-system-react-components.herokuapp.com/?selectedKind=SLDSPopover&selectedStory=Header&full=0&addons=1&stories=1&panelRight=0

PR #1268, opened 2018-02-09
Export canUseDOM correctly to enable focus trap

### LINKED ISSUE #1270: Create unit test for canUseDOM
This broke creating popover issue #1266.

interactivellama:
Fixes issue #1266 and fixes #1270.

### Additional description
~Copied code from React Modal (which is where the original focus trap code came from). I am unsure what changed in browsers. DSR versions from three months were check with this bug and still contained the bug.~

~I still need to add a test to make sure this feature doesn't revert again~

---


interactivellama:
The problem turned out to be an issue with `canUseDOM`. It's export was undefined due to refactoring.

@futuremint I looked at about 10 open source projects that have focus trap--none of them test the trap. There is no way to test it. `node.focus()` is the only way to implement focus change and that doesn't test the trap.

There may be ways to unit test whether the focus trap code detects the correct number of tabbable nodes. We could also add a unit test to test that `canUseDOM` is true. I've created #1270 and will follow up with that.

components/popover/__tests__/popover.browser-test.jsx
```diff
@@ -244,6 +249,35 @@ describe('SLDSPopover', function () {
...
 
+	describe('focus has moved to dialog', function () {
+		const triggerClicked = sinon.spy();
+
+		beforeEach(() => {
+			mountNode = createMountNode({ context: this });
+		});
+
+		afterEach(() => {
+			// destroyMountNode({ wrapper, mountNode });
```
futuremint: If you're not using this should it be removed (along with the whole afterEach callback?)

futuremint:
I don't see where any code in Popover was actually changed here... its just changes to unit tests... am I missing something?

interactivellama:
No, you are not. Turns out it had nothing to do with the Popover code. I forgot to update the PR description. 

It was all a side effect of canUseDOM being undefined and there for the focus/DOM interactions didn't get run. I originally moved code over from React Modal, but when realized that the code focus trap code was not broken. It was just never being run!

Issue #1276, opened 2018-02-14
Document use of AirBnB shape PropType


interactivellama:
Response to #1243.

interactivellama:
This is no longer a valid assertion. `PropType.shape` behavior has changed.

PR #1277, opened 2018-02-14
Document use of AirBnB shape PropType

interactivellama:
Allows objects with undefined values to merge with default values. Very useful with assistiveText and label strings.

Fixes issue #1276


interactivellama:
`shape` from the `PropType` library doesn't allow null values and enforces that all keys are present with non-null values while airbnb doesn't? I will look at the source. I haven't looked at it recently.

interactivellama:
Closing until I can get to the bottom of whether PropType.shape works like AirBnB's shape now. The default behavior looks like it changed.

Issue #1278, opened 2018-02-14
Missing component and proptype descriptions


interactivellama:
https://react.lightningdesignsystem.com/ has many missing descriptions (20+).

This fixes the comment format that `react-docgen` reads which requires `/**` at the start and also adds missing ones.

PR #1279, opened 2018-02-14
Add missing proptype and component descriptions

interactivellama:
Fixes issue #1278 and #1280

We should probably add a test to prevent this by traversing `component-docs.json` after it's built and see if there are any empty strings.

---


interactivellama:
Added #1280 "empty string" test

package.json
```diff
@@ -9,6 +9,7 @@
   },
   "scripts": {
     "build-docs": "./node_modules/.bin/babel-node ./scripts/build-docs.js",
+    "test-docs": "npm run build-docs && $(npm bin)/babel-node ./tests/component-docs-test.js",
```
futuremint: This _also_ builds the docs... and that side effect is then used in `local-qa.sh` and `travis-ci.sh` scripts to build _then_ test the docs.  I propose that this name is not descriptive enough and should be called something like `"build-and-test-docs"`. Or consider instead making `"test-docs"` _only_ actually test the docs, and then change the `DOCGEN` command in the shell scripts to something like `'npm run test-docs && npm run build-docs'` instead of combining it here.

interactivellama: `npm run build-and-test-docs` is fine, but I was looking at it like how we test the JS. We build the JS with Webpack and then run it through Mocha, etc. Also, most contributors do not build the docs that are part of their pull request. We could add that step though. The component docs build is typically done at release.

I guess the difference is that you end up with a file that you can version when you build the docs, but you don't in JS.

interactivellama:
@futuremint Updated.

PR #1299, opened 2018-03-02
Allowing input label of combobox to show as required (*).

### LINKED ISSUE #1295: Combobox: Expose `required` attribute on input text label
I need to use a combobox with a (*) required label.

Combobox uses `<Label>` and that component has a required prop defined.

We should expose it in combobox.

I can add a key to `label` prop called `labelRequired`, that is a boolean.

What do you think?

davidlygagnon:
Fixes #1295

### Additional description

*OUTSIDE SLDS UX PATTERN: Multiple selection Combobox error messages should be placed after pillboxes with an additional `slds-has-error` wrapping div.*

renditiondownload

- ~Adding new key (`labelRequired`) on `label` prop to allow showing (*) on input label. It's already exposed on `<Label/>` but wasn't on `<Combobox>`.~

- Adding a new prop `errorText` that can be passed to show an error message.

- Adding a new prop `aria-describedby` for accessibility. 

screen shot 2018-03-01 at 4 26 42 pm

* [x] While the contributor's branch is checked out, run `npm run local-update` within locally cloned [site repo](https://github.com/salesforce-ux/design-system-react-site) to confirm the site will function correctly at the next release.


components/combobox/__examples__/snapshot/base-label-required.jsx
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable no-console, react/prop-types */
```
interactivellama: I'm not 100% sure, but I think if you don't need the menu open you can just import the "non-snapshot" story into here, since it will look the same.

components/combobox/__tests__/__snapshots__/combobox.snapshot-test.jsx.snap
```diff
@@ -299,7 +300,8 @@ exports[`Base Custom Menu Item Open HTML Snapshot 1`] = `
 "<div class=\\"slds-form-element\\"><label class=\\"slds-form-element__label\\" for=\\"combobox-unique-id\\">Search</label>
   <div class=\\"slds-form-element__control\\">
     <div class=\\"slds-combobox_container\\">
-      <div class=\\"slds-combobox slds-dropdown-trigger slds-dropdown-trigger_click ignore-react-onclickoutside slds-is-open\\" aria-expanded=\\"true\\" aria-haspopup=\\"listbox\\" aria-owns=\\"combobox-unique-id-listbox\\" role=\\"combobox\\">
+      <div class=\\"slds-combobox slds-dropdown-trigger slds-dropdown-trigger_click ignore-react-onclickoutside slds-is-open\\" aria-describedby=\\"HJ4l8ZUm8_G-error-message\\" aria-expanded=\\"true\\" aria-haspopup=\\"listbox\\" aria-owns=\\"combobox-unique-id-listbox\\"
```
interactivellama: `HJ4l8ZUm8_G-error-message` looks like a random assigned ID. this will fail tests.

interactivellama:
Please fix tests.

components/combobox/__examples__/snapshot/base-label-required.jsx
davidlygagnon: Makes sense - I reused the "non-snapshot" version here.

components/combobox/__tests__/__snapshots__/combobox.snapshot-test.jsx.snap
davidlygagnon: Good catch! I updated the test and passed in an ID.

components/combobox/combobox.jsx
```diff
@@ -1191,9 +1217,10 @@ class Combobox extends React.Component {
 				className={classNames('slds-form-element', props.classNameContainer)}
 			>
 				<Label
-					assistiveText={this.props.assistiveText.label}
+					assistiveText={this.props.assistiveText}
```
davidlygagnon: Fixing an unrelated bug here.

davidlygagnon:
@interactivellama This PR is ready for review, when you get a chance. Thanks!

components/combobox/__tests__/__snapshots__/combobox.snapshot-test.jsx.snap
```diff
@@ -360,6 +362,113 @@ exports[`Base Custom Menu Item Open HTML Snapshot 1`] = `
...
+      >
+        <div
+          className="slds-combobox__form-element slds-input-has-icon slds-input-has-icon_right"
+          role="none"
+        >
+          <input
+            aria-activedescendant={null}
+            aria-autocomplete="list"
+            aria-controls="combobox-unique-id-listbox"
+            aria-describedby={undefined}
```
interactivellama: ` aria-describedby="description-unique-id-error-message"` should be here instead despite all the ARIA states up on that `div`.

screen shot 2018-03-12 at 5 16 16 pm

components/combobox/__tests__/__snapshots__/combobox.snapshot-test.jsx.snap
```diff
@@ -360,6 +362,113 @@ exports[`Base Custom Menu Item Open HTML Snapshot 1`] = `
...
+            onInput={undefined}
+            onInvalid={undefined}
+            onKeyDown={[Function]}
+            onKeyPress={undefined}
+            onKeyUp={undefined}
+            onSelect={undefined}
+            onSubmit={undefined}
+            placeholder="Search Salesforce"
+            readOnly={false}
+            required={undefined}
```
interactivellama: This needs to be true for accessibility.

See Input component:
screen shot 2018-03-12 at 5 16 16 pm

components/combobox/__examples__/snapshot/base-open-class-name.jsx
```diff
@@ -45,6 +45,7 @@ class Example extends React.Component {
 		return (
 			<IconSettings iconPath="/assets/icons">
 				<Combobox
+					aria-describedby="described-by-unique-id"
```
interactivellama: What is the reason to add `aria-describedby="described-by-unique-id"` to all snapshots instead of just the "required input" example?

components/combobox/__tests__/combobox.browser-test.jsx
```diff
@@ -211,12 +211,14 @@ describe('SLDSCombobox', function () {
...
 		});
 
-		it('has aria-haspopup, aria-expanded is false when closed, aria-expanded is true when open, ', function () {
+		it('has aria-haspopup, aria-expanded is false when closed, aria-expanded is true when open, has aria-describedby', function () {
 			wrapper = mount(<DemoComponent multiple />, { attachTo: mountNode });
 			const nodes = getNodes({ wrapper });
 			expect(nodes.combobox.node.getAttribute('aria-haspopup')).to.equal(
 				'listbox'
 			);
+			expect(nodes.combobox.node.getAttribute('aria-describedby')).to.not.be
```
interactivellama: This can be removed. `aria-describedby` is a static relationship that can be tested with Jest markup tests.

components/combobox/__tests__/combobox.browser-test.jsx
```diff
@@ -418,5 +420,14 @@ describe('SLDSCombobox', function () {
 					.text()
 			).to.equal('No matches found.');
 		});
+
+		it('Should show the error text', function () {
```
interactivellama: This can removed, too. Markup test is enough.

components/combobox/combobox.jsx
```diff
@@ -132,6 +142,7 @@ const propTypes = {
 	 */
 	labels: PropTypes.shape({
 		label: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
+		labelRequired: PropTypes.boolean,
```
interactivellama: Please be consistent with `Input` and use top level `required` prop.
https://github.com/salesforce/design-system-react/blob/master/components/forms/input/index.jsx#L223

components/component-docs.json
```diff
@@ -2690,6 +2704,13 @@
                     "computed": false
                 }
             },
+            "errorText": {
+                "type": {
+                    "name": "string"
+                },
+                "required": false,
+                "description": "Message to display when the input is in an error state. When this is present, also visually highlights the component as in error."
```
interactivellama: Please add _Tested with snapshot testing._

components/combobox/combobox.jsx
davidlygagnon: ah cool - yeah will do. That's one thought I had when adding that prop. Should have brought it up.

components/combobox/__examples__/snapshot/base-open-class-name.jsx
davidlygagnon: If I don't specify it on all combobox snapshot tests, it will generate a unique ID. That will fail the snapshot tests, right? Or should I get rid of the OR statement in 
```
getErrorId () {
  return `${this.props['aria-describedby'] ||
    this.generatedId}-error-message`;
}
```

interactivellama: See Input's https://github.com/salesforce/design-system-react/blob/master/components/forms/input/index.jsx#L266

davidlygagnon:
@interactivellama I fixed the remaining issues - have a look when you get a chance. Thanks!

components/combobox/combobox.jsx
```diff
@@ -1054,9 +1109,15 @@ class Combobox extends React.Component {
...
+							required={props.required}
 							role="textbox"
 							value={value}
 						/>
+						{props.errorText && (
+							<div id={this.getErrorId()} className="slds-form-element__help">
+								{props.errorText}
+							</div>
+						)}
 						{this.getDialog({
```
interactivellama: When `menuPosition` is `relative` which means the menu's CSS is `position: relative`, the menu needs to be placed after the input.

screen shot 2018-03-13 at 9 30 28 pm

interactivellama:
@davidlygagnon Looks like it will be another day. I need to figure out where the error text goes if the pills are present. If I don't hear anything, I'll go ahead and push and merge the first image below and we can iterate on it in a later release based on an SLDS recommendation.

screen shot 2018-03-13 at 9 58 03 pm
screen shot 2018-03-13 at 9 55 36 pm
screen shot 2018-03-13 at 9 53 22 pm
screen shot 2018-03-13 at 9 43 47 pm

davidlygagnon:
That's a good point. I would think the error message should be the closest to the input as possible and go with image 4, but on the other hand, it feels like the pills are part of the combobox so the error message should come after them, so image 1. I don't have a strong opinion but let's see what SLDS folks think.

interactivellama:
UX PATTERN: Multiple selection Combobox error messages should be placed after pillboxes with an additional slds-has-error wrapping div. See markup for base variant in thread messages.

renditiondownload

PR #1300, opened 2018-03-02
Include ES modules in NPM package

### LINKED ISSUE #1265: Question: Does webpack's tree-shaking work with babel-preset-design-system-react?
I have a question.
In [Getting Started](https://react.lightningdesignsystem.com/getting-started/), it says

> This preset will keep Babel compatible with Design System React and allow ES6 module benefits such as tree-shaking.

but when I see the preset definition,
https://github.com/salesforce/design-system-react/blob/master/preset/index.js

`"modules": false` is not specified in babel-preset-env.
So I just wondered if ES6 modules features like tree-shaking really works using this preset. What do you think?

### LINKED ISSUE #1269: Unable to import the library in Create React App
I came up with this error when trying to start a project from "create-react-app". This has never happened before until I updated the design-system-react to the latest version.

```
./node_modules/@salesforce/design-system-react/components/index.js
Module parse failed: Unexpected token (4:7)
You may need an appropriate loader to handle this file type.
| /* Licensed under BSD 3-Clause - see LICENSE.txt or git.io/sfdc-license */
|
| export SLDSSettings from './settings';
| export Settings from './settings';
|
```

### LINKED ISSUE #1284: Updating package.module to transpiled ES6 modules (CRA compatibility issue)
Create React App now supports ES6 modules (and the transpiling of `node_modules` folder) and is now using `package.modules`. I think this is what is causing #1269. I propose switching `package.modules` to be a transpiled version except for ES6 modules and not the source code.

I'm hesitant to add a third version of this library within the node module, but `main` should be ES5/CJS and `modules` should be ES6, and I'd like to the source code present, especially for internal library users that are using `.babelrc` and want to use `master` branch pull downs in their app, etc.

Another option is remove the CJS version and require ES6 module usage.

This will most likely be 0.9.x and include #1265 's `module.export` fixes.

interactivellama:
Fixes #1284. Fixes #1269. Fixes #1265.

### Additional description
Final solution was ES5 code wrapped in ES6 modules with `package.sideEffects = false`. This allows Webpack 4 to tree shake. 

Allows `package.module` to link to ES6 modules. This has been tested in the following ways:

Uses ES6 modules package in:
Create React App with Webpack 3.

Does tree shaking with
Webpack 4

CRA 2.0 should use Webpack 4 and do the tree shaking.

---

* Run `npm run dist` and inspect files
* Copy/paste `.tmp-npm` into an existing project to see if it works.


interactivellama:
Looks like Karma needs CommonJS babel plugin to run tests now that it's removed from `.babelrc`. Looking into.

interactivellama:
I'm going to step back from this task for a week or so. All that needs to be done is to have the code that goes into Karma/PhantomJS to be CommonJS->UMD and ready for the browser.

There's probably a way to configure webpack with ENV vars or turn Babel settings back into `modules: commonJS` which was just changed in `@salesforce/babel-preset-design-system-react@2.0` to `modules: false` if someone else once to pick this up in order to create compatibility with Create React App.

https://github.com/salesforce/design-system-react/pull/1300/files#diff-a34975409803f73a0edc00817a158877R12

interactivellama:
@davidlygagnon has offered to look into this issue.

interactivellama:
I'm thinking that it may just be easier to run the build/release process with `preset.module` turned on and revert the preset https://github.com/salesforce/design-system-react/blob/master/.babelrc#L2 back to CommonJS, so that folks can decide what kind of module bundling they want to do.

davidlygagnon:
Yeah, that's another option. How would consumer use `preset.module` ? In this case, we won't build a folder with ES6 code anymore?

interactivellama:
We will still build a folder of ES6 modules, we just have to have Babel in ES6 output module when we build them. This can be done by overriding the `module` key during the ES6 build.

Now that I've had it in the back of my mind for weeks, the problem with setting Babel to export as ES6 modules instead of CommonJS is that whatever consumes the post-Babel code has to understand ES6 modules--which is fine for most Webpack configs, but we are kind of dictating that with the preset in v2.0, and I think CommonJS as the default (consumers can always override a preset) which probably be better for the foreseeable future.

interactivellama:
@davidlygagnon I'm transferring this issue to @futuremint since he said he was familiar with Bable transforms, etc.

@futuremint We will need to release a 3.0 version of the babel preset that reverts `module` key back to undefined. Just use the 1.0 for now--I think, and we can bump the preset NPM module when this issue is merged. https://www.npmjs.com/package/@salesforce/babel-preset-design-system-react

futuremint:
@interactivellama ok this is ready to review now.  I tested this in Webpack 4 using the following minimal config:
```
const path = require('path');

module.exports = {
    mode: 'production',
    module: {
        rules: [
            {
                test: /\.jsx?$/,
                include: [path.resolve(__dirname)],
                use: { loader: 'babel-loader' }
            }
        ]
    },
    resolve: {
        extensions: ['.js', '.jsx'],
        alias: { 'design-system-react': path.resolve(__dirname, 'node_modules/@salesforce/design-system-react') }
    },
};
```

And this as the JS file:
```
import { Button, Icon, DataTable, PageHeader } from 'design-system-react';

// All commented out, bundle is: 566 bytes.
Icon(); // just this is 34.2KB 
// Button(); // just this is 211KB
// DataTable(); // just this is 279KB
// PageHeader(); // just this is 221KB
// But all uncommented are 293KB
```

These results required the addition of `sideEffects: false` to the `package.json` file in the npm package. The reason being that all of the re-exports in our big `index.js` file don't automatically get trimmed without it. [Here is the relevant Webpack documentation](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free).

Issue #609, opened 2016-08-17
Make tree fully accessible (implement tree view application role pattern) [34]


interactivellama:
From @SiTaggart, this is the primary source of truth of what show a tree should function:
http://accessibleculture.org/articles/2013/02/not-so-simple-aria-tree-views-and-screen-readers/
http://w3c.github.io/aria/practices/aria-practices.html#TreeView
- Links within nodes and expand buttons have an implicit `tabindex=0` currently. When the tree is made to use the tree application pattern (ARIA role), then all tabbing should be removed ( `tabindex=-1`).
- Current state of the tree is the ARIA tree pattern is not present due to delays in figuring the API for the first nested data controlled component in the library.
- Tree may be a good candidate for the `keyboardNavigable` mixin that is used in MenuDropdown.
- `aria-activedescendant` should be the node id that has focus. This is used when a composite widget is responsible for managing its current active child to reduce the overhead of having all children be focusable. Examples include: multi-level lists, trees, and grids. In some implementations the user agent may use aria-activedescendant to tell assistive technologies that the active descendant has focus. Authors MAY use the aria-activedescendant attribute on the focused descendant of a composite widget; for example, on a textbox descendant of a combo box.
- Having a focus state, should require making the tree component having state and therefore use `React.createClass`.

Additional a11y links for reference (however I'm not sure all the information is accurate):
- http://www.oaa-accessibility.org/examplep/treeview1/
- http://w3c.github.io/aria/practices/aria-practices.html#TreeView
- https://www.w3.org/WAI/GL/wiki/Using_aria-activedescendant_to_allow_changes_in_focus_within_widgets_to_be_communicated_to_Assistive_Technology
  https://www.marcozehe.de/2012/02/06/if-you-use-the-wai-aria-role-application-please-do-so-wisely/

Please review https://github.com/salesforce-ux/design-system-react/blob/master/CONTRIBUTING.md#concepts-and-best-practices for a general overview of concepts used in this library.

interactivellama:
@chriscorwin Are there any updates on this issue?

PR #1323, opened 2018-04-13
Tree keyboard navigation

garygong:
This includes accessibility changes for the Tree component.
- Add role=”presentation” to buttons and anchors
- Add keyboard navigation: up/down to navigate, left/right to open/close
- Add aria-label=”{the label for that node}”

@interactivellama I'd like to get initial impressions on this approach.  Can you let me know what you think?

Fixes https://github.com/salesforce/design-system-react/issues/609

---

* [x] Add year-first date and commit SHA to `last-slds-markup-review` in `package.json` and push.
* [x] Add year-first review date, and commit SHA, `last-accessibility-review`, to `package.json` and push.

### DIFF
--- a/components/tree/index.jsx
+++ b/components/tree/index.jsx
@@ -137,7 +137,7 @@ class Tree extends React.Component {
 					level={0}
 					node={{ nodes: this.props.nodes }}
 					flattenedNodes={this.state.flattenedNodes}
-					selectedNodeIndexes={this.state.selecselectedNodeIndexestedNodes}
+					selectedNodeIndexes={this.state.selectedNodeIndexes}
 					focusedNodeIndex={this.state.focusedNodeIndex}
 					onNodeFocus={this.handleNodeFocus}
 					onClick={this.handleClick}
--- a/components/tree/index.jsx
+++ b/components/tree/index.jsx
@@ -32,8 +32,11 @@ class Tree extends React.Component {
 		this.handleClick = this.handleClick.bind(this);
 		this.handleNodeFocus = this.handleNodeFocus.bind(this);
 		this.state = {
-			flattenedNodes: this.flattenTree({ nodes: this.props.nodes, expanded: true }).slice(1),
-			selectedNodeIndexes: []
+			flattenedNodes: this.flattenTree({
+				nodes: this.props.nodes,
+				expanded: true,
+			}).slice(1),
+			selectedNodeIndexes: [],
 		};
 	}
 
@@ -43,7 +46,10 @@ class Tree extends React.Component {
 
 	componentWillReceiveProps (nextProps) {
 		this.setState({
-			flattenedNodes: this.flattenTree({ nodes: nextProps.nodes, expanded: true }).slice(1)
+			flattenedNodes: this.flattenTree({
+				nodes: nextProps.nodes,
+				expanded: true,
+			}).slice(1),
 		});
 	}
 
@@ -56,7 +62,12 @@ class Tree extends React.Component {
 		if (root.expanded) {
 			for (let index = 0; index < root.nodes.length; index++) {
 				const curNode = root.nodes[index];
-				nodes = nodes.concat(this.flattenTree(curNode, treeIndex ? `${treeIndex}-${index}` : `${index}`));
+				nodes = nodes.concat(
+					this.flattenTree(
+						curNode,
+						treeIndex ? `${treeIndex}-${index}` : `${index}`
+					)
+				);
 			}
 		}
 		return nodes;
@@ -78,13 +89,17 @@ class Tree extends React.Component {
 		// Keep track of the currently selected and focused nodes.
 		let selectedNodeIndexes;
 		if (data.select) {
-			selectedNodeIndexes = this.state.selectedNodeIndexes.concat([data.treeIndex]);
+			selectedNodeIndexes = this.state.selectedNodeIndexes.concat([
+				data.treeIndex,
+			]);
 		} else {
-			selectedNodeIndexes = this.state.selectedNodeIndexes.filter((treeIndex) => treeIndex !== data.treeIndex);
+			selectedNodeIndexes = this.state.selectedNodeIndexes.filter(
+				(treeIndex) => treeIndex !== data.treeIndex
+			);
 		}
 		this.setState({
 			focusedNodeIndex: data.treeIndex,
-			selectedNodeIndexes
+			selectedNodeIndexes,
 		});
 	}
 
--- a/components/tree/index.jsx
+++ b/components/tree/index.jsx
@@ -29,8 +29,11 @@ import { TREE } from '../../utilities/constants';
 class Tree extends React.Component {
 	constructor (props) {
 		super(props);
+		this.handleClick = this.handleClick.bind(this);
+		this.handleNodeFocus = this.handleNodeFocus.bind(this);
 		this.state = {
-			flattenedNodes: this.flattenTree({ nodes: this.props.nodes, expanded: true }).slice(1)
+			flattenedNodes: this.flattenTree({ nodes: this.props.nodes, expanded: true }).slice(1),
+			selectedNodeIndexes: []
 		};
 	}
 
@@ -39,9 +42,9 @@ class Tree extends React.Component {
 	}
 
 	componentWillReceiveProps (nextProps) {
-		this.state = {
+		this.setState({
 			flattenedNodes: this.flattenTree({ nodes: nextProps.nodes, expanded: true }).slice(1)
-		};
+		});
 	}
 
 	// Flattens hierarchical tree structure into a flat array.
@@ -57,6 +60,36 @@ class Tree extends React.Component {
 			}
 		}
 		return nodes;
+	}
+
+	handleClick (event, data, clearSelectedNodes) {
+		// When triggered by a key event, other nodes should be deselected.
+		if (clearSelectedNodes) {
+			this.state.flattenedNodes.forEach((flattenedNode) => {
+				if (flattenedNode.node.selected) {
+					flattenedNode.node.selected = false;
+				}
+			});
+		}
+
+		// Do the click.
+		this.props.onClick(event, data);
+
+		// Keep track of the currently selected and focused nodes.
+		let selectedNodeIndexes;
+		if (data.select) {
+			selectedNodeIndexes = this.state.selectedNodeIndexes.concat([data.treeIndex]);
+		} else {
+			selectedNodeIndexes = this.state.selectedNodeIndexes.filter((treeIndex) => treeIndex !== data.treeIndex);
+		}
+		this.setState({
+			focusedNodeIndex: data.treeIndex,
+			selectedNodeIndexes
+		});
+	}
+
+	handleNodeFocus (event, data) {
+		this.setState({ focusedNodeIndex: data.treeIndex });
 	}
 
 	render () {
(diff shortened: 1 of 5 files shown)


components/tree/index.jsx
```diff
@@ -27,10 +27,38 @@ import { TREE } from '../../utilities/constants';
...
+		};
+	}
+
 	componentWillMount () {
 		checkProps(TREE, this.props);
 	}
 
+	componentWillReceiveProps () {
+		this.state = {
+			flattenedNodes: this.flattenTree({ nodes: this.props.nodes, expanded: true }).slice(1)
```
garygong: We re-flatten the nodes here on every props change, which may not be ideal.

components/tree/private/branch.jsx
```diff
@@ -155,13 +240,20 @@ const renderBranch = (children, props) => {
...
 
-	// TODO: Remove tabbing from anchor tag AND button / add tabIndex={-1} when keyboard navigation is present.
 	return (
 		<li
 			id={props.htmlId}
 			role="treeitem"
 			aria-level={props.level}
 			aria-expanded={isExpanded ? 'true' : 'false'}
+			aria-label={props.node.nodes && props.node.nodes.length > 0 ? props.node.label : null}
+			tabIndex="0"
```
interactivellama: All items need to be `tabIndex="-1"` except for the active/focused one, so that there's only one tab stop

components/tree/private/item.jsx
```diff
@@ -39,18 +42,92 @@ const handleClick = (event, props) => {
...
+			select: true,
+			treeIndex: flattenedNode.treeIndex,
+		});
+	} else {
+		// Focus is on this node but it's not selected, so select it
+		handleClick(event, props);
+	}
+};
+
+const handleKeyDownLeft = (event, props) => {
```
interactivellama: `left` should close its branch

components/tree/index.jsx
```diff
@@ -27,10 +27,38 @@ import { TREE } from '../../utilities/constants';
...
 	}
 
+	componentWillReceiveProps (nextProps) {
+		this.state = {
+			flattenedNodes: this.flattenTree({ nodes: nextProps.nodes, expanded: true }).slice(1)
+		};
+	}
+
+	// Flattens hierarchical tree structure into a flat array.
+	flattenTree (root, treeIndex = '') {
```
interactivellama: This seems to be performant enough with small trees. There's a large dataset story, but it's already some slowness with select/unselect, so I think it's the DOM update and not any kind of data within the component.

components/tree/private/branch.jsx
```diff
@@ -69,6 +73,87 @@ const handleScroll = (event, props) => {
...
+				select: true,
+				treeIndex: props.flattenedNodes[index].treeIndex,
+			});
+		}
+	}
+};
+
+const handleKeyDown = (event, props) => {
+	mapKeyEventCallbacks(event, {
+		callbacks: {
```
interactivellama: We'll need an `enter` public callback and all the multi-select stuff per the SLDS spec here.

interactivellama:
@arodenbeck Could you look over the Tree component? Keyboard navigation and auto-select have been added in this PR. https://design-system-react-co-pr-1323.herokuapp.com/?selectedKind=SLDSTree&selectedStory=Base&full=0&addons=0&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

It's still a WIP and needs to handle multi-select for one thing. I'm interested on your take on what happens when you `tab` into a tree? Does tabbing into a Tree select the first item, since it has the default `tabIndex`?

garygong:
Yeah, I saw that too.  I'll move to those key actions.  Plus the Ctrl-space to select, too.

SiTaggart:
You can leave multiple select on re-enter until you press and arrow key. Just focus me still on the last one I was focued on.

arodenbeck:
Right arrow should also move focus to the first child node, if a branch is already expanded. This is kind of the opposite of left arrow focusing the parent of an expanded branch.

SiTaggart:
Yup, Absolutely @arodenbeck

components/tree/private/item.jsx
garygong: In a comment below, @arodenbeck implied that this behavior is ok.  (Left arrow selects the parent node.)

components/tree/private/branch.jsx
garygong: I tied `Enter` to the item click callback.  Is that what you mean?

garygong:
@interactivellama I implemented the keyboard navigation behavior that @SiTaggart and @arodenbeck pointed us to in this commit: https://github.com/salesforce/design-system-react/pull/1323/commits/8da59a8ad5ac5887ff7126fc93d36cfeca2b02c3

Can you guys take a look and let me know what you think?  Thanks!

components/tree/private/branch.jsx
```diff
@@ -182,12 +182,12 @@ const handleKeyDownEnter = (event, props) => {
 const handleKeyDown = (event, props) => {
 	mapKeyEventCallbacks(event, {
 		callbacks: {
```
arodenbeck: @garygong I'm having some strange behavior around tabindexes. If the selection is the first item, things act as expected. If you select any other item however, the first item maintains its tabindex, adding an extra tab stop to the tree.

garygong: Thanks for finding this.  It turns out that this was caused by a typo in the code!  It's fixed now and the new link is here: https://design-system-react-co-pr-1323.herokuapp.com/?selectedKind=SLDSTree&selectedStory=Base&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

Let me know if you see anything else.  Thanks!

arodenbeck: I don't think I understand the Highlighted Search variant. Should the tree get filtered down to only nodes which match the search? When I type in the search box, my focus is moved to the tree as soon as I press a key. I was able to type "nut", but I had to shift tab back to the input twice to add letters.

components/tree/private/branch.jsx
interactivellama: I confirmed that click and enter should be the same callback.

components/tree/private/branch.jsx
```diff
@@ -155,13 +296,25 @@ const renderBranch = (children, props) => {
...
+			aria-label={
+				props.node.nodes && props.node.nodes.length > 0
+					? props.node.label
+					: null
+			}
+			tabIndex={getTabIndex(props)}
+			onKeyDown={(event) => handleKeyDown(event, props)}
+			onFocus={(event) => handleFocus(event, props)}
+			ref={(component) => {
+				if (component && isFocused) {
```
interactivellama: probably need another flag like `isTreeFocused` here.

interactivellama:
**Update from accessibility team:** Unfortunately, the SLDS spec on multi-select is not fully fleshed out. The site lists what is the ARIA spec, but doesn't take into consideration Ctrl vs Cmd (Win/Mac) and OS-X standard keyboard comments like Cmd+space (spotlight). The team recommends removing support for keyboard-based multiple select.

garygong:
Thanks for looking.  I'll fix the issues found, and also remove the multi-select keyboard actions.

components/tree/index.jsx
```diff
@@ -53,6 +53,16 @@ class Tree extends React.Component {
 		});
 	}
 
+	shouldComponentUpdate (nextProps, nextState) {
```
garygong: I know this is kind of weird, but this is the simplest solution that I could think of.  The problem that I was running into was that when clicking on a node, the blur is triggered and it would re-render the tree; as a result, the click event would never be triggered.

garygong:
@interactivellama @arodenbeck Please check out my latest commit, which fixes the issues that you guys found.  Thanks!  https://github.com/salesforce/design-system-react/pull/1323/commits/d2348f7797bb46b8bb9a53a30f5d07019d4e6083

Updated sample here: https://design-system-react-co-pr-1323.herokuapp.com/?selectedKind=SLDSTree&selectedStory=Base&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

components/tree/index.jsx
```diff
@@ -27,10 +27,97 @@ import { TREE } from '../../utilities/constants';
...
+	componentWillReceiveProps (nextProps) {
+		this.setState({
+			flattenedNodes: this.flattenTree({
+				nodes: nextProps.nodes,
+				expanded: true,
+			}).slice(1),
+		});
+	}
+
+	shouldComponentUpdate (nextProps, nextState) {
```
garygong: I know this is kind of weird, but this is the simplest solution that I could think of. The problem that I was running into was that when clicking on a node, the blur is triggered and it would re-render the tree; as a result, the click event would never be triggered.

SiTaggart:
Interaction seems pretty good, seems to re-render the entire tree on every key press though

garygong:
Thanks for taking a look.  Yeah, the implementation relies on re-rendering the tree on key actions.

components/tree/private/branch.jsx
```diff
@@ -69,6 +72,118 @@ const handleScroll = (event, props) => {
...
+};
+
+const handleKeyDownLeft = (event, props) => {
+	if (props.node.expanded) {
+		handleExpandClick(event, props);
+	} else {
+		const nodes = props.flattenedNodes.map(
+			(flattenedNode) => flattenedNode.node
+		);
+		const index = nodes.indexOf(props.parent);
```
interactivellama: I found one more thing. We need to close the parent on handleKeyDownLeft if on an item.

> Left arrow key on an end child node, *collapses the group* and moves :focus and aria-selected to the parent `treeitem`.

components/tree/private/item.jsx
```diff
@@ -39,32 +42,140 @@ const handleClick = (event, props) => {
...
+				treeIndex: flattenedNode.treeIndex,
+			},
+			true
+		);
+	}
+};
+
+const handleKeyDownLeft = (event, props) => {
+	const nodes = props.flattenedNodes.map((flattenedNode) => flattenedNode.node);
+	const index = nodes.indexOf(props.parent);
```
interactivellama: Whoops. This needs to be on the item not a branch. We need to close the parent on handleKeyDownLeft if on an item.

> Left arrow key on an end child node, *collapses the group* and moves :focus and aria-selected to the parent `treeitem`.

components/tree/private/item.jsx
```diff
@@ -39,32 +42,140 @@ const handleClick = (event, props) => {
 	}
 };
 
+const findNextNode = (flattenedNodes, node) => {
+	const nodes = flattenedNodes.map((flattenedNode) => flattenedNode.node);
+	const index = nodes.indexOf(node);
+	return flattenedNodes[(index + 1) % flattenedNodes.length];
```
interactivellama: Nice use of modulus here.

interactivellama:
Re-rendering is the only I can think of to move focus if one is using internal state and DOM refs.

@garygong This is looking great! This needs tests. Something like [Combobox's SelectedListBox](https://github.com/salesforce/design-system-react/blob/master/components/combobox/__tests__/combobox.browser-test.jsx#L260)

@SiTaggart Are you approving this from a a11y standpoint?

components/tree/private/item.jsx
garygong: I added this behavior in this commit.  And also added the tests!  https://github.com/salesforce/design-system-react/pull/1323/commits/1a36d887eecb89beaab50c543c49968bfd230799

PR #1326, opened 2018-04-16
Allow importing lookup DefaultSectionDivider

Vintik:
I do know that lookup is currently deprecated, but we still use it and I've spent a full day trying to migrate one lookup to a combobox.

In the meantime this DefaultSectionDivider used to be exposed in es6 modules, but not in commonjs. It'd be a whole world easier if we just exposed it in both until the lookup component is actually killed. Refactoring is not part of the scope of upgrading DSR for me.

### DIFF
--- a/components/index.js
+++ b/components/index.js
@@ -126,6 +126,12 @@ export InputIcon from './icon/input-icon';
 
 export SLDSLookup from './lookup';
 export Lookup from './lookup';
+export SLDSLookupDefaultFooter from './lookup/menu/default-footer';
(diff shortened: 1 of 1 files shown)


interactivellama:
@Vintik I've added all Lookup's parts and prefixed. Does this look good to you?

Vintik:
Yeah looks fine to me, I thought about doing that but decided to keep the change minimal. I guess it doesn't really matter much since the entire component is deprecated and nobody should be creating new ones at this point.

PR #1334, opened 2018-04-24
Require `iconCategory` if `iconName` is set 

### LINKED ISSUE #342: Require `iconCategory` if `iconName` is set in `<Button>` and `<Dropdown>`
References [this comment](https://github.com/salesforce-ux/design-system-react/pull/341#issuecomment-227736582) on #341

Here are a few packages that could help accomplish this:
[github.com/evcohen/react-proptype-conditional-require](https://github.com/evcohen/react-proptype-conditional-require)
[github.com/thejameskyle/react-required-if](https://github.com/thejameskyle/react-required-if)

davidlygagnon:
Fixes #342 

### Additional description
Adding react-required-if to allow conditional required prop (https://github.com/jamiebuilds/react-required-if)

Fixing bug where we want to have a required prop `iconCategory` when `iconName` is set.
---


components/button/index.jsx
```diff
@@ -66,13 +67,10 @@ const Button = createReactClass({
...
 		 * Name of the icon category. Visit <a href="http://www.lightningdesignsystem.com/resources/icons">Lightning Design System Icons</a> to reference icon categories.
 		 */
-		iconCategory: PropTypes.oneOf([
-			'action',
-			'custom',
-			'doctype',
-			'standard',
-			'utility',
-		]),
+		iconCategory: requiredIf(
```
davidlygagnon: There's a default prop set for `iconCategory` so this will really only work (meaning the warning message will only show) if we get rid of the default prop. Do we still want to do this here? Probably a good idea to add `requiredIf` package but maybe this prop doesn't need a conditional given it has a default prop.

interactivellama: One option would be to move the default into the component (private), so that it is backwards compatible, but we change the public API so folks get the warning.

interactivellama:
This looks good. I'll pull it down and test it tomorrow.

davidlygagnon:
Great thanks!
I just noticed that there are warnings appearing when running the snapshot tests ```Warning: Failed prop type: The prop `iconCategory` is marked as required in `SLDSButton`, but its value is `undefined`.```, which is expected now whenever `iconName` is set. 

But should I fix all usages of `<Button>` where `iconName` is passed without an `iconCategory` in our code? Some components use `<Button iconName>` without `iconCategory`, e.g.: panel.jsx. Otherwise, people will start seeing the warning when consuming DSR components.

interactivellama:
Yes.

interactivellama:
There are some more examples that need updated. Search for `iconName="down"`

davidlygagnon:
Cool thanks - going to look into it today.

components/menu-dropdown/menu-dropdown.jsx
```diff
@@ -184,13 +185,10 @@ const MenuDropdown = createReactClass({
...
 		 * Name of the icon category. Visit <a href="http://www.lightningdesignsystem.com/resources/icons">Lightning Design System Icons</a> to reference icon categories.
 		 */
-		iconCategory: PropTypes.oneOf([
-			'action',
-			'custom',
-			'doctype',
-			'standard',
-			'utility',
-		]),
+		iconCategory: requiredIf(
```
davidlygagnon: 👍

PR #1344, opened 2018-04-30
Make hint an optional field for data table actions

Vintik:
Still have no idea what a hint is. Couldn't really find anything about hints at https://www.lightningdesignsystem.com/components/buttons/#hint as specified across multiple locations in docs and code. Is this a deprecated thing?

Either way it's messing with the styling of the dropdown icon for me, and there's no way to override it.

### DIFF
--- a/components/data-table/row-actions.jsx
+++ b/components/data-table/row-actions.jsx
@@ -47,6 +47,10 @@ const DataTableRowActions = createReactClass({
 		 */
 		item: PropTypes.object,
 		/**
+		 * Disable hint styling which changes the color of the dropdown svg on hover over.
+		 */
+		noHint: PropTypes.bool,
+		/**
 		 * Triggered when an item is selected.
 		 */
 		onAction: PropTypes.func,
@@ -59,6 +63,7 @@ const DataTableRowActions = createReactClass({
 	getDefaultProps () {
 		return {
 			assistiveText: 'Actions',
+			noHint: false,
 		};
 	},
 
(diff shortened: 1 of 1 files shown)


interactivellama:
It's a "hover hint" and is when the color of the down arrow on the button changes from gray to black on hover of the table row.

interactivellama:
Please add prop to comments so that it's listed on the website for folks to see [like so](https://react.lightningdesignsystem.com/components/data-tables/#prop-row-actions-assistiveText). Probably something like `/* Changes the color of SVGs in child nodes due to the mouse hover on ancestor nodes. */` if that makes better sense.

If true is the default, we need to make it `noHint` so we can be consistent [with not having defaults that are true](https://github.com/salesforce/design-system-react/blob/master/docs/codebase-overview.md#boolean-prop-prefix)

components/data-table/row-actions.jsx
```diff
@@ -46,6 +46,10 @@ const DataTableRowActions = createReactClass({
 		 * `DataTable` row item
 		 */
 		item: PropTypes.object,
+		/**
+		 * Disable hint styling which changes the color of the dropdown svg on hover over.
```
Vintik: @interactivellama didn't want to worry the user about child/ancestor elements since it's all self contained within this component. The ancestor is the dropdown and the child is the icon.

Vintik: Does this automatically add to the doc site? I have no idea how this works.

interactivellama: Yes, `npm run build-docs` will create `components/component-docs.json` which is ingested by the doc site.

interactivellama: The ancestor is the table row. Hover over the row on the left side and watch the color change of the button on the right side.

vintik:
Lol what happened here :)

interactivellama:
I don't know. 😃

Issue #1236, opened 2018-01-23
Modals: No header example does not aria-hide the rest of the page


arodenbeck:
The 2nd example, no header, does not appear to hide the rest of the page. This allows a screen reader user to perceive parts of the page that should not be functional.

arodenbeck:
The aria-labeledby does not seem to be properly labeling  this modal either. I'm guessing the referenced Id should be in the header, which this variant does not have. An aria-label could be used here instead, or possibly pass in a label.

PR #1347, opened 2018-05-03
Modal accessibility fixes

garygong:
- aria-hidden="true" isn't applied when there are more than one Modals
- use passed-in assistiveText when there is no header

Fixes https://github.com/salesforce/design-system-react/issues/1236

---


components/modal/__examples__/header-footer.jsx
```diff
@@ -23,58 +23,63 @@ const Example = createReactClass({
 	},
 
 	render () {
+		const modal = this.state.noHeaderIsOpen ? (
```
garygong: `aria-hidden=true` isn't applied when there is more than one modal rendered.  This is actually a bug in react-modal, which is fixed in v2.0.6.  But when I tried updating to it, I saw some test failures (something about onClose handler not being triggered) that I didn't spend too much time looking into.  The right thing to do is to probably fix up our code so that we can use the newer version of react-modal.

components/modal/index.jsx
```diff
@@ -29,6 +29,10 @@ const propTypes = {
 	 * Vertical alignment of Modal.
 	 */
 	align: PropTypes.oneOf(['top', 'center']),
+	/**
+	 * Assistive text for the modal. If not provided, `title` is used.
+	 */
+	assistiveText: PropTypes.string,
```
interactivellama: Let's make this an object with a key of `dialog-label`, so that it can be future proof. We may have some more keys to add when we upgrade.

interactivellama:
@arodenbeck Could you look at the modal `aria-label` when you have a chance. https://design-system-react-co-pr-1347.herokuapp.com/?selectedKind=SLDSModal&selectedStory=Small%20no%20header&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

components/modal/index.jsx
garygong: I updated it with `dialogLabel`in camel case because I saw that's how other instances of `assistiveText` were defined.  

Also, I noticed that we could put `closeButtonAssistiveText` in there as well, but I'm not sure what the process is for relocating a prop?

interactivellama: That's a great idea. There's a `checkProps` function that can run at component mount that will provide a console error in dev mode.

Take a look at [tooltip](https://github.com/salesforce/design-system-react/blob/master/components/popover/check-props.js#L27). This example is the move from `isInline` to `position=relative`.

Then for the actual code, you remove the old prop from the `propTypes` which removes it from the doc site, but you still access it with a `text = newProp || oldProp` privately inside the `render` function, so that it doesn't actually broke anyone's code. Then, we'll officially break it in the future after folks have been given time to change to the new API. @davidlygagnon is working on the something pretty similar right now, too.

interactivellama:
I believe "Modal no header" is added as the aria-label of the modal, since the dev or designer would have requested no visible heading.

arodenbeck:
Looks like everything in issue 1236 has been fixed.

interactivellama:
Looks great! @garygong

PR #1350, opened 2018-05-04
Warn about future changes to onChange callback [WIP]

### LINKED ISSUE #273: Audit components: event callbacks should pass syntheticEvent, then customData object
Suggestion from @tweettypography 

Make all custom event callbacks have two properties `handleEvent(customData, syntheticEvent)`. The first prop would be an object with whatever is being passed back from the app - updated values, etc - and the second is simply the original triggering event. 

TO DOS
- Change all components to this pattern
- Update docs

interactivellama:
Fixes #273

### Additional description
* Removed breaking change label. This isn't one and just moves components under `/forms` that are not deprecated, aliases have been added to make backwards compatible. The current code is NOT a breaking change, but it will update the documentation to promote use of `/components/checkbox` instead.
* The audit showed that checkbox and inline edit's `onChange` parameters are wrong (along with deprecated lookup's onChange, onBlur, onFocus, but we aren't updating Lookup). 
* **DataTable's `onChange` is also `(selectedArrayOfItems, event)`. I'm going think about that one a little more.** I'm thinking `onRowChange` or `onRowSelect` would be the best backwards compatible option right now.
* SLDS has moved their form components out from under the form folder, so a breaking change in event parameters was a good time to move the `checkbox` component out of the forms folder. That way, consumers can upgrade at their leisure--unless they are using the CommonJS named imports. 
* Right now, CommonJS and `components/forms/checkbox` users (that is everyone) will get the warning, but nothing will break.
* Went ahead and converted checkbox to a ES6 class and removed a lodash function.
* There was a comment in all the site-stories that mentioned `forms-checkbox` and that has been removed.
* The first commit "Align event prop parameters" just adds from empty objects to a few components as the second parameter `data` object to existing event callbacks, just to guard from being `undefined` if accessed. This should be standardized.

Deprecation warning for InlineEdit
inline edit warning

Checkbox warning:
screen shot 2018-05-18 at 5 04 56 pm

---


components/checkbox/check-props.js
```diff
@@ -0,0 +1,36 @@
...
+if (process.env.NODE_ENV !== 'production') {
+	checkProps = function (COMPONENT, props) {
+		deprecatedEventParameter(
+			COMPONENT,
+			{
+				oldEventParameterOrder: props.oldEventParameterOrder,
+				propAsString: 'onChange',
+				propAsValue: props.onChange,
+			},
+			'`components/forms/checkbox` is deprecated. `components/checkbox` should be used. When this path update is made `onChange` event parameters will change from `onChange(value, event, { value } to `onChange(event, { value }). Please update your event parameters when you change paths.` If you are using the CommonJS named import, `Checkbox` events will break at v1.0 and this warning will be present until then. Please review https://github.com/salesforce/design-system-react/releases when you upgrade.'
```
davidlygagnon: Minor: missing `)` in `onChange(value, event, { value } to ...`

components/checkbox/check-props.js
```diff
@@ -0,0 +1,36 @@
...
+		);
+
+		if (props.variant === 'toggle' && props.indeterminate === true) {
```
davidlygagnon: Minor : `=== true` not necessary?

Issue #1356, opened 2018-05-07
Tree keyboard navigation bug


interactivellama:
this doesn’t work:

tab into tree, tab out, tab back in, click a new item
it doesn’t select the item

PR #1358, opened 2018-05-08
Fix tree keyboard navigation

garygong:
This fixes an issue with the tree keyboard navigation:
- Tab into tree
- Tab out
- Tab back in
- Click a new item
- The item isn't selected

This also has an additional fix for when hitting left on a branch.  If the branch is already collapsed, it should collapse the parent branch and focus the parent node.

Fixes https://github.com/salesforce/design-system-react/issues/1356

### Additional description

---


components/tree/index.jsx
```diff
@@ -57,7 +58,7 @@ class Tree extends React.Component {
 		// There is no need to render when blurring a node because focus is either:
 		//  - outside of the tree, or
 		//  - focused on another node in the tree, which triggers its own render
-		if (this.state.treeHasFocus && !nextState.treeHasFocus) {
+		if (!nextState.treeHasFocus) {
```
garygong: it turns out that this is all we need, and before it was causing the issue described in https://github.com/salesforce/design-system-react/issues/1356

components/tree/index.jsx
```diff
@@ -118,6 +119,11 @@ class Tree extends React.Component {
 		this.setState({ treeHasFocus: false });
 	}
 
+	handleExpandClick (event, data) {
```
garygong: One thing that we need to do is to trigger a re-render as mentioned in the comment above.  An good thing is that now I see the branch loading placeholders, whereas before I didn't see them.  I guess we weren't triggering the re-renders for the before.

components/tree/private/branch.jsx
```diff
@@ -138,6 +138,11 @@ const handleKeyDownLeft = (event, props) => {
 		);
 		const index = nodes.indexOf(props.parent);
 		if (index !== -1) {
+			props.onExpandClick(event, {
```
garygong: Hitting left on a branch that's already collapsed should collapse the parent branch and focus the parent node, like how it works for collapsing an item.

interactivellama:
Good find on the loading stencils

Issue #1374, opened 2018-05-11
Progress Indicator Style/Color Updates


kwilloughby:
The Progress Indicator needs to be updated to fix the following issues:

- The blue Tooltip should always use regular Tooltip text (i.e. no bold, no italic)
- The blue Tooltip should always be blue even if an error occurs on a step
- The error icon (red triangle) needs to be updated to the new error icon (red ban sign) - this might automatically change when we update to Summer ’18 (SLDS 2.6.0)

** Note: If an error occurs a second error icon should be placed to the left of the buttons (e.g. Cancel, Save) and the Error Popover should appear indicating there are errors. These additional items are NOT part of this component. This note was included for visibility purposes. **

interactivellama:
<img width="389" alt="screen shot 2018-07-05 at 3 30 46 pm" src="https://user-images.githubusercontent.com/1290832/42346486-6a7d0612-8068-11e8-9128-7a19198780cb.png">

PR #1376, opened 2018-05-13
Fixed problem with progress indicator/assistive-text

### LINKED ISSUE #1238: Progress Indicators: Needs updates to the assistive-text
This component has similar issues to Tooltips, see #1228. From this examaple, it seems like there is visual indication as to whether a step has been completed. Is there also visual indication as to which step is currently displayed? If so, I would like to see aria-selected added to the current step.

kchan902:
… completed steps. Also added in assistive text messages for completed and disabled steps

Fixes #1238 and #1374

### Additional description
* Adds `assistiveText` props
  * ~`activeStep`: Label for the active step. The default is `Active Step`~
  * `completedStep`: Label for a completed step. The default is `Completed Step`
  * `disabledStep`: Label for disabled step. The default is `Disabled Step`
  * `errorStep`: Label for a step with an error. The default is `Error Step`
  * `step`: Label for a step. It will be typically followed by the number of the step such as "Step 1".
* Update error icon
* add `aria-current=step`

---

* [x] Request a review of the deployed Heroku app by the Salesforce UX Accessibility Team.


kchan902:
@interactivellama Could you please review this pull request when you have a chance? I am not able to assign it for review. Thank you!

interactivellama:
@arodenbeck Please review https://design-system-react-co-pr-1376.herokuapp.com/ and confirm it fixes the issue you created in January #1238.

arodenbeck:
@kchan902 Thanks for adding the aria-current="step", that does work well. I wonder if we might address some of the extra wording that the screen reader is saying. For example, now that we have aria-current="step", we no longer need "Active Step" to be added in as "slds-assistive-text". Currently the screen reader is saying "Step 5: Active Step - Button - Current Step", which is a bit much.

I also wonder about how the new props will work for the assistive text labels? It looks like there is one called step, which is the label. The notes then say the number would come after the label? I would think we would want it more like "Step 1: Email Attributes".

Can we tweak the assistive text for the disabled and error states as well? Currently it is hard to differentiate status information from the actual label of the step. SLDS separates the label from the status portion with a " - ", which makes better reading. This way you would get "Step 2: Email Attributes - Disabled". It also eliminates the word "step" as part of the status.

The markup for disabled steps needs to be tweaked. Currently there is a <span> that has a role="button", and a tabindex="0". Since the intent here is to make it act like a button, it should just be a <button>, to match the markup of the other steps.

The progress bar does not need a tabindex="0". Since it isn't used to perform an action, it doesn't need to be in the tab order.

interactivellama:
Summarizing @arodenbeck 
* Remove "Active Step"
* Separate disabled and error text from label name with "-". Remove "step" from "Disabled Step" or "Error Step"
* Make label format: ""Step 2: Email Attributes - Disabled"
* disabled step should be `a` with `role=button`
* Remove `tabindex="0"` from progress bar node.

arodenbeck:
@kchan902 Looking good. I did notice a few cosmetic things which might not exist in production, so let me know if these are irrelevant.
* The step names seem to have the - separator as part of their label. This causes a problem for steps which do not have a Disabled or Error tag at the end. Example "undefined 4: tooltip label #4 -" There is a trailing -
* The variant "Completed progress", has "Finished step" rather than Completed. It looked like the messaging can be configured by the person using the component, so was this just defined as an example?
* Several of the variants use a label of "This is custom text in the assistive text key" for the first step. I know it is up to the consuming dev to define the names of the steps, but that's a strange message.
* This might be big enough to stand as its own issue, and I'll be glad to create one separate if you think it necessary. Each step triggers a tooltip on focus with a <span> of the label of the step. I think that was part of the initial design, but that functionality is unnecessary, and doesn't match the current SLDS version. This tooltip style functionality should be removed.

kchan902:
@arodenbeck for points 2 and 3, yes it was defined as an example. for point 4, please create a separate issue. thanks!

Issue #1378, opened 2018-05-14
Tree: Show proper aria-label value when label is a node


garygong:
Render a tree using a node value as one of its node values, for example:
```
label: <span>Fruits</span>,
type: 'branch',
id: 2,
```

This results in an incorrect aria-label value:
![screen shot 2018-05-14 at 10 22 31 am](https://user-images.githubusercontent.com/7608505/40013196-b87aa7b4-5761-11e8-955c-2cdea66153ae.png)

I suggest adding `assistiveText` to the node, like this:
```
assistiveText: 'Fruits',
label: <span>Fruits</span>,
type: 'branch',
id: 2,
```

Issue #1377, opened 2018-05-14
Tree is not re-rendered on props changes


garygong:
- Render a tree
- Trigger a props change
- The tree should be re-rendered based on the props change, but it doesn't.

PR #1379, opened 2018-05-14
Tree bug fixes introduced by keyboard navigation

garygong:
This fixes two Tree bugs: 
- Tree is not rendered on props changes https://github.com/salesforce/design-system-react/issues/1377
- Show proper aria-label value when label is a node https://github.com/salesforce/design-system-react/issues/1378

---

### DIFF
--- a/components/tree/index.jsx
+++ b/components/tree/index.jsx
@@ -29,9 +29,9 @@ import { TREE } from '../../utilities/constants';
 class Tree extends React.Component {
 	constructor (props) {
 		super(props);
-		this.handleClick = this.handleClick.bind(this);
+		this.handleSelect = this.handleSelect.bind(this);
 		this.handleNodeBlur = this.handleNodeBlur.bind(this);
-		this.handleExpandClick = this.handleExpandClick.bind(this);
+		this.handleExpand = this.handleExpand.bind(this);
 		this.state = {
 			flattenedNodes: this.flattenTree({
 				nodes: this.props.nodes,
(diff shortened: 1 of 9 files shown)


components/tree/index.jsx
```diff
@@ -54,16 +54,6 @@ class Tree extends React.Component {
 		});
 	}
 
-	shouldComponentUpdate (nextProps, nextState) {
```
garygong: It turns out that it's simpler to avoid the react lifecycle complete by moving treeHasFocus out of state.

interactivellama: What I'm seeing here is something like `treeHasFocusDueToClick`? Would that be more accurate? I'll go ahead and push that variable name if you agree.

garygong: Yes that's right, go ahead.  Thanks!

components/tree/private/branch.jsx
```diff
@@ -36,22 +36,22 @@ import mapKeyEventCallbacks from '../../../utilities/key-callbacks';
...
 	}
 };
 
-const handleClick = (event, props) => {
+const handleSelect = (event, props) => {
 	EventUtil.trap(event);
-	if (isFunction(props.onClick)) {
-		props.onClick(event, {
+	if (isFunction(props.onSelect)) {
+		props.onSelect(event, {
```
interactivellama: @garygong Removing this line makes the expand work.

components/tree/private/branch.jsx
```diff
@@ -173,7 +172,7 @@ const handleKeyDown = (event, props) => {
 };
 
 const handleFocus = (event, props) => {
-	if (!props.focusedNodeIndex) {
+	if (!props.focusedNodeIndex && event.target === event.currentTarget) {
```
garygong: This selects the node upon first focus, but only do it if focusing the `<li>`.  When focusing the button, we allow it to go through so that the branch expand handler works.

PR #1387, opened 2018-05-18
Consolidate assistiveText props under one object [App Launcher, Popover, Modal]

garygong:
This PR consolidate assistive text props under one `assistiveText` prop for the following components:

- App Launcher
- Popover
- Modal

More components to come!

---

### DIFF
--- a/components/app-launcher/index.jsx
+++ b/components/app-launcher/index.jsx
@@ -69,6 +69,12 @@ const AppLauncher = createReactClass({
 	// ### Prop Types
 	propTypes: {
 		/**
+		 * Assistive text for the app launcher.
+		 */
+		assistiveText: PropTypes.shape({
+			trigger: PropTypes.string, // Assistive text for app launcher icon
+		}),
+		/**
 		 * One or more `<AppLauncherSection />`s each containing one or more `<AppLauncherTile />`s
 		 */
 		children: PropTypes.node.isRequired,
@@ -105,10 +111,6 @@ const AppLauncher = createReactClass({
 		 */
 		title: PropTypes.string,
 		/**
-		 * Assistive text for app launcher icon
-		 */
-		triggerAssistiveText: PropTypes.string,
-		/**
 		 * This is typically the name of the cloud or application
 		 */
 		triggerName: PropTypes.node,
@@ -120,7 +122,7 @@ const AppLauncher = createReactClass({
 
 	getDefaultProps () {
 		return {
-			triggerAssistiveText: 'Open App Launcher',
+			assistiveText: {},
 			title: 'App Launcher',
 		};
 	},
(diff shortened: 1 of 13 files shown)


components/app-launcher/index.jsx
```diff
@@ -120,7 +122,7 @@ const AppLauncher = createReactClass({
 
 	getDefaultProps () {
 		return {
-			triggerAssistiveText: 'Open App Launcher',
+			assistiveText: {},
```
interactivellama: We will need to provide the default for `this.props.assistiveText.trigger` here and do an object merge with a `...` spread.

[This one](https://github.com/salesforce/design-system-react/blob/master/components/alert/index.jsx#L80) is done with `assign`, but use native if possible.  https://github.com/salesforce/design-system-react/blob/master/components/alert/index.jsx#L120

interactivellama: We will also want to list out an object keys in the comment description. https://github.com/salesforce/design-system-react/blob/master/components/alert/index.jsx#L20

interactivellama:
Preferred output on doc site:
screen shot 2018-05-18 at 10 04 52 am

https://react.lightningdesignsystem.com/components/alerts/

components/app-launcher/index.jsx
garygong: Good catch on the comment description.

As for the defautProps value, I don't think there's a way in React to have defaultProps.trigger plus have `this.assistiveText.trigger || this.props.triggerAssistiveText` work because it will always take the default prop value.  

We could still do the object assign, but omit the new prop (`trigger`) and then add it back when we take away the deprecated prop (`this.props.triggerAssistiveText`) for good. What do you think?

garygong: Something like in this commit: https://github.com/salesforce/design-system-react/pull/1387/commits/6d3648f5ec607d26486228dfb43c6accf2835fb5 
Let me know what you think.

interactivellama: I think we want to do something like:

```
const assistiveText = { defaultProps.assistiveText, ...this.props.assistiveText}
let assistiveTextTrigger = this.props.triggerAssistiveText ? this.props.triggerAssistiveText :  assistiveText.trigger

[use assistiveTextTrigger]
```
If they are using the old _and_ the new API, then I'm OK with the old API overriding things--they are still in the driver's seat. It's not following the spec, and should get a console warning about the old API anyway. If you'd like to add something to the `deprecationWarning` stating that `this.props.triggerAssistiveText` will take precedence and be used if it exists that might not be a bad idea.

garygong: Yeah, I agree.  Also, I didn't know about the object spread syntax!  Good call!

Issue #1211, opened 2018-01-09
Popover looks wrong when changing position


george-norris-salesforce:
[![https://gyazo.com/0a3eee7bcdcedec09728142ca74c75ce](https://i.gyazo.com/0a3eee7bcdcedec09728142ca74c75ce.gif)](https://gyazo.com/0a3eee7bcdcedec09728142ca74c75ce)

• When there is room on screen, popover resizes and does not look good (the pointer is located at the top, and does not keep locked at the target button).

Would be nice to have an option to have pointer anchor to one position and not change position, or to have the popover alignment change (always keeping the pointer locked at the target) when there is limited room.

interactivellama:
There is a `hasStaticAlignment` https://design-system-react.herokuapp.com/components/popovers/#prop-hasStaticAlignment option if you'd prefer it to go off the page. Looks like the description isn't making it into the docs though. I'll take a look at the doc parsing.

Flipping the arrow is still a feature that needs to be added if you want to take a crack at it. It would require getting the info from PopperJS that the Popover had flipped and re-rendering.

george-norris-salesforce:
I can take a crack at the arrow flipping.

It appears that `hasStaticAlignment` does not work correctly though..
[![https://gyazo.com/b4738d371e871aa8badbc98bf58c225f](https://i.gyazo.com/b4738d371e871aa8badbc98bf58c225f.gif)](https://gyazo.com/b4738d371e871aa8badbc98bf58c225f)

interactivellama:
@Vintik I've assigned this to @garygong. He will be working on this in the next two weeks.

interactivellama:
@garygong @Vintik worked this into #1404. Apologies for the back and forth! 😞

interactivellama:
Fixes #1211

PR #1304, opened 2018-03-09
Allow nubbin to change position when popover flips

george-norris-salesforce:
Fixes #1211.

Allows the nubbin to change position when a popover flips

Before

[https://gyazo.com/d4a444546e61100ef2b532cd7e1eb255](https://gyazo.com/d4a444546e61100ef2b532cd7e1eb255)

After:

[https://gyazo.com/d419d7f31b8bb6c628722a326a233d75](https://gyazo.com/d419d7f31b8bb6c628722a326a233d75)
---

### Pull Request Review checklist (do not remove)

* [ ] Review the appropriate Storybook stories. Open [http://localhost:9001/](http://localhost:9001/).
* [ ] Review tests are passing in the browser. Open [http://localhost:8001/](http://localhost:8001/).
* [ ] Review markup conforms to [SLDS](https://www.lightningdesignsystem.com/) by looking at snapshot strings.
* [ ] Add year-first date and commit SHA to `last-slds-markup-review` in `package.json` and push.
* [ ] Request a review of the deployed Heroku app by the Salesforce UX Accessibility Team.
* [ ] Add year-first review date, and commit SHA, `last-accessibility-review`, to `package.json` and push.
* [ ] While the contributor's branch is checked out, run `npm run local-update` within locally cloned [site repo](https://github.com/salesforce-ux/design-system-react-site) to confirm the site will function correctly at the next release.

PR #1404, opened 2018-06-02
Make Dialog positioning more robust

Vintik:
This pull request is meant to make dialog positioning more robust and move responsibility closer to the Dialog component instead of having the parent/consumer be responsible for the dialog logic.

In my eyes, the responsibility of the parent should be only what gets rendered in the Dialog. Not how the dialog renders. It can use props to change that on a case by case basis. Currently the coupling between parent and Dialog is very tight and needs to be loosened.

Our internal utility Dialog component now receives a `hasNubbin` which will add a nubbin. The placement of the nubbin will be determined from the overall `align`. Nubbins can only be added onto `position="absolute"` or `position="overflowBoundaryElement"`

Take a look at any of the alignments for PopoverTooltips:
https://design-system-react-components.herokuapp.com/?selectedKind=SLDSPopoverTooltip&selectedStory=Alignment%20%28Button%29&full=0&addons=0&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

Big things to figure out:
- Getting snapshot tests to pass

- Due to reasons for trapping focus, the `slds-popover` class is required to be a child of `Dialog` and not on dialog, otherwise it breaks. I'd much rather prefer to pass these higher level classes to dialog and not have special divs that try to control the internals of Dialog.

- After these changes the nubbin always points at the center of the element instead of being on the side of the element. Unless we require the user to explicitly provide the exact nubbin position, I think this is the much more reasonable default because it works consistently in many more cases. Otherwise, what would a top left nubbin position on a top left align mean? The nubbin would point up even though the reference element is below the dialog. 

- In a few of the parents I change the wrapper from `display: inline` to `display: inline-block`. Can it be done? Is this a breaking change? We want the "container" to take up the full height in whatever the user renders it in. Otherwise the nubbin will always point at the wrong part.

- preventOverflow. Pretty sure this has never worked properly, but if we add offsets this may not work or the addition of offsets may need to move to a different time in the popper cycle. Either way I think this part would be out of scope unless this actually works for anybody currently using Dialog.
---


components/menu-dropdown/__docs__/storybook-stories.jsx
```diff
@@ -12,6 +12,7 @@ import { DropdownNubbinPositions } from '../../menu-dropdown/menu-dropdown';
 import List from '../../utilities/menu-list';
 import Button from '../../button';
 import Trigger from '../../menu-dropdown/button-trigger';
+import Popover from '../../popover';
```
Vintik: Ignore this file. Just making more complex test cases.

components/utilities/dialog/__docs__/storybook-stories.jsx
```diff
@@ -0,0 +1,31 @@
+// TODO - we should make stories to see everything on one page
```
Vintik: Ignore this. Was looking into having a story of all things that use dialog to see all cases on one page, but didn't have the time for this.

components/utilities/dialog/index.jsx
```diff
@@ -18,12 +18,13 @@ import Portal from './portal';
...
 import KEYS from '../../../utilities/key-code';
 import DOMElementFocus from '../../../utilities/dom-element-focus';
-import { mapPropToPopperPlacement } from '../../../utilities/dialog-helpers';
+import { getNubbinMargins, getNubbinClassName, mapPropToPopperPlacement } from '../../../utilities/dialog-helpers';
 
 import { DIALOG } from '../../../utilities/constants';
 
 // #### Dialog doesn't pass down <IconSettings> context so repassing it here.
 import IconSettings from '../../icon-settings';
+// import { , getNubbinClassName } from '../../utilities/dialog-helpers';
```
Vintik: Will remove

components/utilities/dialog/index.jsx
```diff
@@ -188,23 +193,13 @@ const Dialog = createReactClass({
...
-			this.props.position === 'relative'
-		) {
+		if (this.props.position === 'absolute' || this.props.position === 'relative') {
 			this.handleOpen();
 		}
 	},
 
 	componentDidUpdate (prevProps, prevState) {
-		if (
-			this.state.triggerPopperJS === true &&
```
Vintik: Ignore these formatting changes, i'll change back. I can't read the code as linted. It hurts my eyes, but I'll change back before committing.

components/utilities/dialog/index.jsx
```diff
@@ -242,11 +237,17 @@ const Dialog = createReactClass({
...
 		}
 
-		const propOffsets = this.getPropOffsetsInPixels(this.props.offset);
 		const { position } = popperData.offsets.popper;
-		const left = `${popperData.offsets.popper.left + propOffsets.horizontal}px`;
-		const top = `${popperData.offsets.popper.top + propOffsets.vertical}px`;
-		return { ...popperData.style, left, top, position };
+		const propOffsets = this.getPropOffsetsInPixels(this.props.offset);
+
+		// FIXME before merge - gotta rename from margin to offset
```
Vintik: Got to address fixme and rename to getNubbinOffsets

components/utilities/dialog/index.jsx
```diff
@@ -370,26 +369,21 @@ const Dialog = createReactClass({
...
-		if (
-			this.props.position === 'absolute' ||
-			this.props.position === 'overflowBoundaryElement'
-		) {
-			style = this.getPopperStyles();
+		const style = {};
+
+		if (this.props.position === 'absolute' || this.props.position === 'overflowBoundaryElement') {
+			Object.assign(style, {
+				outline: 0,
```
Vintik: I don't think the outline: 0 is actually needed. Will remove.

components/utilities/dialog/index.jsx
```diff
@@ -412,15 +427,18 @@ const Dialog = createReactClass({
...
 					) || undefined
 				}
+				id={this.props.id}
 				style={style}
 				onKeyDown={this.handleKeyDown}
 				onMouseEnter={this.props.onMouseEnter}
 				onMouseLeave={this.props.onMouseLeave}
 				{...this.props.containerProps}
 				ref={this.setDialogContent}
+				role={this.props.variant}
```
Vintik: Not sure how to address this properly.

components/popover/popover.jsx
```diff
@@ -478,7 +475,6 @@ const Popover = createReactClass({
 					aria-describedby={`${this.getId()}-dialog-body`}
 					className={classNames(
 						'slds-popover',
```
Vintik: This is incredibly annoying since it can't be moved up to the dialog level or at least I'm not sure how to go about it since it's difficult to follow all the logic of trapping the focus in the popover and following this up and down through the components, life cycle methods, and refs.

interactivellama: I think it's easier to follow from a comparing the HTML nodes here to SLDS if you pass in the whole popover.

components/utilities/dialog/index.jsx
interactivellama: This is fine.

interactivellama:
The output definitly looks nicer:
screen shot 2018-06-07 at 12 35 52 am. From an architecture, we should be as "smart" as we can as long as folks can override things to be flexible.

On the subject of trapping focus, the [first tabbable node in the dialog contents should receive focus](https://www.lightningdesignsystem.com/components/popovers/). I made the dialog `div` get focus, because I knew it would always exist and I wouldn't need to do a re-render to focus via component ref. I'm open to changing, but it's tricky because there's a mix of DOM query and React rendering.

I think it's reasonable to ask folks to be inline block if they want to absolutely align something to it.

I'm not sure what you mean by `preventOverflow` not working. If you are referring to the position flip/swap. [This PR](https://github.com/salesforce/design-system-react/pull/1304) will make the nubbin swap positions if you want to incorporate it into this PR.

Something is off with the nubbin like z-index or something:
screen shot 2018-06-07 at 12 22 12 am

components/utilities/dialog/index.jsx
Vintik: Yeah that's what I was thinking, but that's only if the direct variant -> role mapping ends up insufficient.

interactivellama:
@Vintik, @davidlygagnon will be working on Popover's nubbin flip #1211 probably next week and be in the Dialog codebase also. Just FYI.

components/popover/popover.jsx
Vintik: For now leaving it as is, but after thinking about this some more, I think that I disagree, the Dialog is the Popover. You should be able to render an empty popover. Why should it be `div div.popover #contents` when it can just be `div.popover #contents`. It's a bit annoying because we're trying to unify all these components under the concept of a "Dialog" but SLDS limits us in this fashion. Leaving it as is, because I'm already rushed for time as is.

Vintik:
I was thinking about adding it in this week if @davidlygagnon is okay with that. I finally have some time cleared up and we need these dialog changes merged within the next 2 weeks. About to have sprint planning. Hopefully this remains my top priority.

davidlygagnon:
Sounds good @Vintik . I'll take another issue that @interactivellama assigned to me. Thanks!

components/utilities/dialog/index.jsx
```diff
@@ -294,12 +314,10 @@ const Dialog = createReactClass({
 	},
 
 	handleOpen () {
-		const scopedElement = this.dialogContent;
-
-		if (this.props.variant === 'popover' && scopedElement) {
+		if (this.props.variant === 'popover' && this.dialogContent) {
 			DOMElementFocus.storeActiveElement();
 			DOMElementFocus.setupScopedFocus({
-				ancestorElement: scopedElement.querySelector('.slds-popover'),
```
Vintik: Was kind of weird that we used to do a search for slds-popover, even though slds-popover is a class we control. Now since the Dialog no longer receives a div child that it's tightly bound to, and the Dialog is of a type of popover, we can just use the dialogContent reference directly. 

If we want to be a bit defensive, we can do the scan if the class isn't on dialogContent, but then there'd be a bug anyways and the tests should catch that. I don't think this is an appropriate time for defensive checks

components/utilities/dialog/index.jsx
```diff
@@ -414,15 +435,19 @@ const Dialog = createReactClass({
 							[`${this.props.outsideClickIgnoreClass}`]:
 								this.props.position === 'overflowBoundaryElement',
 						},
+						this.props.hasNubbin && getNubbinClassName(this.props.align),
 						this.props.contentsClassName
 					) || undefined
 				}
+				id={this.props.id}
```
Vintik: Should be removed. This should be passed as containerProps instead.

utilities/dialog-helpers.js
```diff
@@ -47,74 +45,81 @@ const getNubbinClassName = (align) =>
...
+const DISTANCE_OFFSET = 1.5; // 'rem'
+const NUBBIN_SIZE = 1; // 'rem'
+const ROTATED_HEIGHT = NUBBIN_SIZE / Math.sqrt(2); // 'rem'
+/*
+*
+*
+*
+*
+*/
+// FIXME - still need to account for border shadow of 2px. probably only needs to be added to the rotated height.
```
Vintik: Not sure this was ever accounted for previously.

Vintik:
Not exactly sure how I resolve the conflicts? :)

interactivellama:
Just pull down master. These images are unrelated to what you are doing.

components/popover/popover.jsx
```diff
@@ -570,7 +549,7 @@ const Popover = createReactClass({
 
 		this.renderOverlay(this.getIsOpen());
 
-		const containerStyles = { display: 'inline' };
+		const containerStyles = { display: 'inline-block' }; // ATTN FIXME - can we make this change??!
```
interactivellama: I think it will be OK. `slds-button` should already be inline block

```
/**
 * @summary This neutralizes all the base styles making it look like a text link
 *
 * @name base
 * @selector .slds-button
 * @restrict button, a, span
 * @variant
 */
.slds-button {
  position: relative;
  display: inline-block;
  padding: 0;
```

components/utilities/dialog/index.jsx
interactivellama: The focusing helpers were originally taken from jQuery UI, so DOM query snuck in.

components/utilities/dialog/index.jsx
```diff
@@ -372,13 +389,16 @@ const Dialog = createReactClass({
 	},
 
 	render () {
-		let style = {};
+		const style = {};
```
interactivellama: See master branch. This needs to `let`. `Object.assign` was removed and spread operator was put in it's place.

components/utilities/dialog/index.jsx
```diff
@@ -372,13 +389,16 @@ const Dialog = createReactClass({
...
 
 		if (
 			this.props.position === 'absolute' ||
 			this.props.position === 'overflowBoundaryElement'
 		) {
-			style = this.getPopperStyles();
+			Object.assign(style, {
```
interactivellama: use spread here.

interactivellama:
Tooltips are not being read on focus of trigger anymore
Your PR:
https://design-system-react-co-pr-1404.herokuapp.com/?selectedKind=SLDSPopoverTooltip&selectedStory=Base&full=0&addons=1&stories=1&panelRight=0&addonPanel=%40storybook%2Faddon-a11y%2Fpanel

Master branch:
http://design-system-react-components.herokuapp.com/?selectedKind=SLDSPopoverTooltip&selectedStory=Base&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

Cmd F5 will enable Voiceover on a Mac. It should read the tooltip when the button is focused. This will need to be fixed. The button and the dialog should be connected by the aria attributes which will cause the button to read the tool contents when focued.

interactivellama:
We'll need to make sure we mention the offset change in the release notes. A checkProps would be nice, but it wouldn't go away after the offset numbers were updated.

Vintik:
@interactivellama So finally got time again :) lol it's been a crazy couple of months. 

Looked into it and realize now why tooltip trap focus doesn't work. It was never supposed to work in the first place and according to the original design would probably be a bug. Frankly, have no idea how it's trapping focus on master branch.

https://github.com/salesforce/design-system-react/blob/master/components/utilities/dialog/index.jsx#L170

components/utilities/dialog/index.jsx
Vintik: Still uses a dom reference, there's just no need to query, there's only one possible element it can be.

components/popover/popover.jsx
Vintik: Oh i need to remove the comment

interactivellama:
Thanks @Vintik. Awesome work!

interactivellama:
Fixes #1211

PR #1406, opened 2018-06-03
Update tree example to be hashmap in order to promote immutability

### LINKED ISSUE #1399: Update tree example to be immutable
The Current tree example uses `forceUpdate()` since the nodes are not immutable and are updated without creating a new node for the updated node. This task is to remove `forceUpdate()` and the alternative `_.deepClone`. `deepClone` is the giant happen to make things immutable, but I'd like to promote best practices in your example and plan to use [Normalizr](https://github.com/paularmstrong/normalizr/blob/master/docs/api.md#schema) (e.g.- "hash table") with object spread to inject the one node that needs to be changed with cloning the whole nested tree. 

On a related note, every tree nodes gets rendered due to `shortId.generate()` being called on every render of every node.

Investigation into using a normalized or flattened tree (array of nodes with ids of children) led to finding a bug that prevents `getNodes` from being used correctly

interactivellama:
Fixes #1399

### Additional description
Creates a best practices example for the doc site using a flattened tree (hashmap) that enables better immutability in Flux/Redux state management of the Tree component. Backwards compatible.

[Nested objects shouldn't be used in Redux](https://medium.com/farmdrop/using-normalizr-js-in-a-redux-store-96ab33991369). `deepClone` is an option, but causes a re-render of the whole tree.

The answer appears to be to use `normalizr` to flatten the tree into an array. Then use `getNodes` to walk the React render tree and get the data from the flattened array. This makes the array immutable, but keeps the object references that did not change intact which should prevent React from re-rendering.

* Adds a `log` utility that will route logs to the correct environment such as Storybook actions, `console.log` (on the docs site), or a custom function that we could extend in the future
* Tree: Add additional prop types
* Update Tree sample nodes to be objects. This allows normalizr/getNodes to work correctly
* Add Storybook actions
* Clean up Tree comments
* Tree: Update tests
* Adds more precise tests to getNodes
* Remove warning by adding heading
* Add test to check number of times getNodes to called (to prevent not calling getNodes in the future)
* Update Tree proptypes: allow ids to be numbers and headings to nodes
* Remove Tree as prototype
* Tree: Add normalizr to flatten tree example
* Clarify internal callback args by using options object

---


components/index.js
```diff
@@ -231,3 +231,6 @@ export UtilityIcon from './utilities/utility-icon';
 
 export SLDSTree from './tree';
 export Tree from './tree';
+
+// Library utilities
+export log from '../utilities/log';
```
garygong: what is this export used for?

components/tree/__docs__/storybook-stories.jsx
```diff
@@ -1,164 +1,69 @@
...
-					onClick={this.handleClick}
-					onScroll={this.handleScroll}
-					searchTerm={this.state.searchTerm}
-					{...this.props}
-				/>
-			</div>
-		);
-	},
-});
+// View the generated hash map
```
garygong: should we clean up the commented `console.logs`?

components/index.js
interactivellama: Adds a `log` utility that will route logs to the correct environment such as Storybook actions, `console.log` (on the docs site), or a custom function that we could extend in the future. 

My mistake that was missing from the description. This should enable us to use the same stories in Storybook as in the doc website and place the complexity of the multi-environment output within this public function.

components/tree/__docs__/storybook-stories.jsx
interactivellama: I'd like to keep it, so folks can debug `normalizr`'s output more easily.

Issue #1403, opened 2018-06-02
RowActions (within DataTable) should support all properties of SLDSMenuDropdown


danperry-salesforce:
Specifically I want to implement the menuPosition and length attributes from SLDSMenuDropdown but I will add all the attributes for completeness.
- I will work on this
- Its related to the following work item: W-4505052

interactivellama:
Thank you for wanting to help out the library. I think this is a great idea!

I would like to minimize duplicate documentation if possible--that is listing all the same props in two different components--it's prone to become out of date. Would you be open to doing something like the [Filter/Popover](https://github.com/salesforce/design-system-react/blob/master/components/filter/index.jsx#L104) prop? It is similar to a Higher-Order Component in that it takes a Popover, grabs the props, and merges them with the default props provided by the parent.

This should:
* give you access to all the drop down props
* Allow autocomplete to work in IDEs for free
* limit the API surface area

danperry-salesforce:
@interactivellama my plan was to do something similar ala: 
```javascript
	{
		/**
		 * `Dropdown` properties.
		 * This will not override equivalent properties on this component (ie. onAction or options)
		 * See `Dropdown`
		 */
		 dropdownProps: PropTypes.shape(Dropdown.propTypes)
	}
```
and merge them in from there.  This way we wouldn't need to construct a component.  Is that ok?

danperry-salesforce:
Here is how I was merging them (still haven't tested):
```javascript
	// ### Render
	render () {
		// i18n
		let defaultDropdownProps = {
			align: 'right',
			buttonClassName: 'slds-button--icon-x-small',
			buttonVariant: 'icon',
			iconCategory: 'utility',
			iconName: 'down',
			iconSize: 'small',
			iconVariant: 'border-filled'
		};

		let firstLevelProps = {
			assistiveText: this.props.assistiveText,
			className: this.props.className,
			options: this.props.options,
			//hint needs to remain undefined or merge will overwrite 'hint' in props.dropdownProps
			hint: isUndefined(this.props.noHint) ? undefined : !this.props.noHint,
			id: this.props.id,
			onSelect: this.handleSelect,
		};

		let dropdownProps = merge({}, defaultDropdownProps, this.props.dropdownProps, firstLevelProps);

		return (
			/* eslint-disable jsx-a11y/no-static-element-interactions */
			<td
				className=""
				data-label="Actions"
				onClick={this.handleClick}
				style={{ width: '3.25rem' }}
			>
[...]

PR #1413, opened 2018-06-07
Added dropdown option to DataTableRowActions

danperry-salesforce:
Fixes https://github.com/salesforce/design-system-react/issues/1403
Added dropdown option to DataTableRowActions that takes MenuDropdown component as a property to allow support for all dropdown properties by overriding defaults

---

### DIFF
--- a/components/data-table/row-actions.jsx
+++ b/components/data-table/row-actions.jsx
@@ -60,7 +60,7 @@ const DataTableRowActions = createReactClass({
 		options: PropTypes.array,
 		/**
 		 * A [Dropdown](http://react.lightningdesignsystem.com/components/dropdown-menus/) component. The props from this drop will be merged and override any default props.
-		 * **Note:** onAction will not be overridden, both `DropDown`'s onSelect and onAction will be called with appropriate parameters
+		 * **Note:** onAction will not be overridden, both `DropDown`'s onSelect(dropDownActionOption) and onAction(rowItem, dropdownActionOption) will be called with appropriate parameters
 		 */
 		dropdown: PropTypes.node,
 	},
--- a/components/data-table/row-actions.jsx
+++ b/components/data-table/row-actions.jsx
@@ -60,7 +60,7 @@ const DataTableRowActions = createReactClass({
 		/**
 		 * `Dropdown` options. See `Dropdown`.
 		 */
-		options: PropTypes.array.isRequired,
+		options: PropTypes.array,
 		/**
 		 * A `Dropdown` component. The props from this drop will be merged and override any default props.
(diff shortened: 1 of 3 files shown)


components/data-table/row-actions.jsx
```diff
@@ -75,11 +84,36 @@ const DataTableRowActions = createReactClass({
...
 		// i18n
+		const defaultDropdownProps = {
+			align: 'right',
+			buttonClassName: 'slds-button--icon-x-small',
+			buttonVariant: 'icon',
+			iconCategory: 'utility',
+			iconName: 'down',
+			iconSize: 'small',
+			iconVariant: 'border-filled',
+			assistiveText: this.props.assistiveText,
```
garygong: do you think that we should deprecate these overlapping props, since we can have a real `Dropdown` now?  what do you think @interactivellama ?

garygong:
look good me to, can you take a quick look @interactivellama ?  thanks!

components/data-table/row-actions.jsx
interactivellama: Yes, we should. I didn't want to require that of someone's first pull request, so I didn't say anything.

jamesward:
@danperry-salesforce You now have an invite to this org: https://github.com/salesforce

Once you accept, get the CLA bot to be happy by going to: https://cla.salesforce.com/status/salesforce/design-system-react/pull/1413

components/data-table/row-actions.jsx
```diff
@@ -57,13 +57,19 @@ const DataTableRowActions = createReactClass({
 		/**
 		 * `Dropdown` options. See `Dropdown`.
 		 */
-		options: PropTypes.array.isRequired,
+		options: PropTypes.array,
+		/**
+		 * A `Dropdown` component. The props from this drop will be merged and override any default props.
+		 * **Note:** onAction will not be overridden, both `DropDown`'s onSelect and onAction will be called with appropriate parameters
```
interactivellama: We probably need a console error here.  See https://github.com/salesforce/design-system-react/blob/master/components/input/check-props.js#L61 for an example of `onlyOneOfProperties`.

danperry-salesforce: Do we need this?  onAction wasn't required before, and they behave differently.  onAction calls with the row item as a param whereas onSelect only provides the dropdown action.

interactivellama: Great question! Let's replace the note part with sharing the difference between the two callbacks and document this topic. List out the parameters and such. I think that will also help down the road with maintainability, when someone (like myself) says, why do we need both of these--without looking at the source code.

PR #1421, opened 2018-06-12
Document component composition with prop spread

interactivellama:
### Additional description
Replaces "Child component decorator pattern" as the preferred composition pattern and best practice of this library. Please review for clarity, since this is so abstract.

This explains why https://github.com/salesforce/design-system-react/pull/1413 does what it does. I'm open to a better name for this pattern, too!


interactivellama:
@danperry-salesforce Does this makes sense and explain the pattern well? Please comment with any suggestions.

PR #1426, opened 2018-06-19
Make Slider a controlled component

### LINKED ISSUE #1398: Slider: Allow value to be set
Currently Slider doesn't set the `value` of the input, and therefore cannot be a controlled component. This means it just doesn't work.

The following should work, but since the value attribute of `input` doesn't ever change `onChange` is never fired.

```
const DemoSlider = createReactClass({
	displayName: 'DemoSlider',

	getInitialState () {
		return {
			value: this.props.value,
		};
	},

	handleChange (event, { value }) {
		this.setState({ value: parseInt(value, 10) });
	},

	render () {
		return (
			<Slider
				{...this.props}
				value={this.state.value}
				onChange={this.handleChange}
			/>
		);
	},
});

const Example = createReactClass({
	displayName: 'SliderExample',

	render () {
		return (
			<section>
				<ol>
					<li className="slds-p-bottom--large">
						<h1 className="slds-text-title_caps slds-p-vertical--medium">
							1. Base Input with label
						</h1>
						<DemoSlider id="base-id" label="My Label" value={50} />
					</li>
				</ol>
			</section>
		);
	},
});

export default Example;

```

DanFerro:
Now sets the value attribute from props.value.

Fixes #1398

---


interactivellama:
@garygong This was the question I was asking about earlier that was whether a change to a component was a breaking change if the component did not work as intended in the first place.

interactivellama:
Moving Review request to @futuremint if he is available this week.

futuremint:
I'm available and I'll review this today.

components/slider/index.jsx
```diff
@@ -140,13 +139,13 @@ class Slider extends React.Component {
 
 	handleChange = (event) => {
 		if (isFunction(this.props.onChange)) {
-			this.props.onChange(event, { value: event.target.value });
+			this.props.onChange(event, { value: parseInt(event.target.value, 10) });
```
futuremint: This could pose a problem should a user want to make a slider for floats (a slider between 0 and 1 for example).  I'm not sure we want to have `parseInt` here because we can't assume.

components/utilities/dialog/index.jsx
```diff
@@ -400,7 +400,7 @@ const Dialog = createReactClass({
 		}
 
 		if (this.props.style) {
-			style = Object.assign(style, this.props.style);
+			style = { ...style, ...this.props.style };
```
futuremint: 👍

utilities/warning/url-exists.js
```diff
@@ -13,32 +13,43 @@ if (process.env.NODE_ENV !== 'production') {
...
-				const additionalComment = comment ? ` ${comment}` : '';
-				/* eslint-disable max-len */
-				warning(
-					!url,
-					`The icon asset was not found at ${url}. Make sure the path to the icon asset is correct. You can set the icon path by importing the IconSettings component, \`<IconSettings iconPath=[/assets/icons]>\` from \`components/iconSettings\`, and wrap that component around your entire app or around individual components using icons. If you are using the \`<Icon>\` component, you can also pass the url to \`this.props.path\`.${additionalComment}`
-				);
-				/* eslint-enable max-len */
-				hasWarned[`${control}-path`] = !!url;
-			}
+	const warn = (control, url, comment) => (res) => {
```
futuremint: Is there a reason to add support for `fetch` here? Does this address a maintenance issue in the backlog or something else?

components/slider/index.jsx
interactivellama: value return text, but we expect number be passed in. I'll look into using `Number()`.

interactivellama:
@futuremint 
* Fetch commit removed after pull from upstream.
* `parseInt` replaced with `Number()`

PR #1430, opened 2018-06-24
Create new SLDSTimePicker component and deprecate current one

### LINKED ISSUE #1169: Combobox Read Only variant: Add disabled property
Add a boolean disabled property to the combobox setting the HTML disabled attribute on the input tag.

@interactivellama Added
Add `readOnlySingleInputDisabled` prop to allow disabling input when in read-only single input variant.

### LINKED ISSUE #1232: Timepickers: Update to use Combobox
Currently timepicker functions like a combobox, but screen readers only understand it to be an input.

davidlygagnon:
Fixes #1232
Fixes #1169

1. Deprecating `components/time-picker` in favor of new component `components/timepicker` which uses combobox internally.
2. ~Adding `iconRight` prop to combobox to allow displaying an icon to the right of the input.~ Adding `input` prop to combobox to expose input props.
3. Adding `readOnlySingleInputDisabled` prop to allow disabling input when in read-only single input variant.

### Additional description

New component is sunsetting props that were present in the old component and not available anymore in the new one.

New component shows a checkmark for selected item. This is new compared to the old one but inline with SLDS Timepickers [SLDS Timepickers](http://www.lightningdesignsystem.com/components/timepicker/)

---


interactivellama:
Should we do a `props.input[node]` prop for Combobox so we re-use the input props API instead of `iconRight`?

davidlygagnon:
You mean allow passing in `/components/input` as prop instead of the iconRight?

davidlygagnon:
@interactivellama When you get a chance, this PR is ready for review. Thanks!

interactivellama:
The new TimePicker should be a controlled component only and not switch to uncontrolled. It should also be autocomplete/inline-listbox type of Combobox, so that typing in the input filters similar to `comboboxFilterAndLimit` the selected options like in Combobox.

davidlygagnon:
Updates on this: 

The TimePicker is now a variant of `inline-listbox`.

That means, users can now type into the input text box and time filtering happens.

newtimepicker

I also went ahead and deprecated `onDateChange` in favor of using `onChange` and `onInputChange` to stay aligned with Combobox. `onChange` is called when there's a keyboard event - `onSelect` happens when a menu item is selected (which would be the same as onDateChange). This change was also a result from getting `invalid date` when typing in the input (for example, when starting to type `1` for `1:00`), `onDateChange` was returning invalid dates on the first input.

Let me know what you think @interactivellama  , thanks.

Vintik:
Probably already thought of question, but should you actually see 12:00 when you type in 2:00 ? Seems annoying to see 12:00 since that's obviously not what I'd be typing.

davidlygagnon:
@Vintik Yeah makes sense. We can easily fix that or even consider providing a filter override function. I think this refactor is on hold for the moment.

interactivellama:
It's been de-prioritized, since June because no one has complained about it and @davidlygagnon has been working on other Analytics Cloud needs (which is what he should be work on contributing).

Feel free to pick it up @Vintik .

The two things I can think of left to do are:
        * add `onSubmit`
        * remove `predefinedOptionsOnly` so you can do other options.

interactivellama:
Some kind of `time-filter.js` could be added the example in place of the `combobox` default.

PR #1442, opened 2018-07-05
Update Tree branch and item to use new SLDS HTML tags

vernak2539:
### Additional description

When the Tree component is used for left hand navigation, which is quite often, when the tree gets big and expands to the whole space, when users hover over a item/branch, the link URL is displayed in the lower left hand corner of the browser window (Chrome).

We've seen that if the link element is long or the tree item is short user's can miss the tree item at the bottom of the screen. 

To combat this, we are submitting this PR which removes the `href` attr from the Tree's branch and item. 

After checking [W3](https://www.w3.org/TR/2016/REC-html51-20161101/textlevel-semantics.html#the-a-element), we've found the following:

>If the a element has no href attribute, then the element represents a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element’s contents.

---


vernak2539:
@interactivellama does this look good?

interactivellama:
Looks good! Thanks for the update.

PR #1457, opened 2018-07-13
Prefer object spread eslint rule

interactivellama:
Add eslint rule to prefer spread operator over Object.assign(). This only gets flagged though when the first parameter is `{}`.


futuremint:
FYI Travis is failing because there is still an `Object.assign` in https://github.com/salesforce/design-system-react/blob/master/webpack.config.dist.js#L12 😄

interactivellama:
Not production code. Merging. Tests pass now.

PR #1468, opened 2018-07-20
Support developing with React 16, and as a peer dependency

### LINKED ISSUE #1282: Make library compatible with React 16
While keeping React 15.5 for dev and tests until time can be allotted to update test framework.

There are React 16 users out there and they are not reporting issues, but that's anecdotal.

futuremint:
Fixes #1282. Note: no code changes to any components were needed for this.

This changes the dev dependency to React 16 as well as the version of ReactDOM, Enzyme, and Test Utils that are being used. 

Unit tests and a few older storybook examples have been updated where needed to run with the newer version of React and Enzyme.  They also run 50% faster now in the browser (5s vs. 11s in master).

### Additional description

---

* [ ] Update docs where any React version is specifically mentioned.
* [ ] Run `npm run local-update` within locally cloned [site repo](https://github.com/salesforce-ux/design-system-react-site) to confirm the site will function correctly at the next release.


package.json
```diff
@@ -113,8 +113,8 @@
 	},
 	"peerDependencies": {
 		"@salesforce-ux/design-system": "^2.6.0",
-		"react": ">=15.4.1 <16",
-		"react-dom": ">=15.4.1 <16"
+		"react": ">=16",
```
interactivellama: This should be this way in the `devDependencies`, but are you sure we shouldn't include 15 since you only changed test code? We could also bump the minor version, too, just to make sure it's clear to consumers even if it's not a breaking change except npm and introduce `npm install` warnings. We have communicated to teams that we were going to do this after March, but I'm thinking we need to re-iterate it before releasing. What do you think?

interactivellama:
We will want to update any docs that mention a version. https://github.com/salesforce/design-system-react#getting-started

package.json
futuremint: Yes, thanks for catching this.  I mean to put it back to including 15.4.1 and above... I'll get this done this afternoon or tomorrow.

futuremint:
Thanks, I forgot about the docs, I'll get those updated too.

PR #1496, opened 2018-08-09
Update react-doc-gen version

interactivellama:
Helps with reading component in #1495


interactivellama:
No production code changes.

PR #1526, opened 2018-09-01
Fix table column widths to actually apply

### LINKED ISSUE #1525: FixedLayout DataTableColumn widths don't actually apply
Repro:

Go to the documentation or storybooks and observe that the <th> never gets the width from the column. Only the <td> gets it. Not sure how this ever worked in any previous version.

From my observations the <td> doesn't actually need the width and the column width is mainly determined by the <th>

Most my application's fixed tables also just lost any sort auto-default sizing when upgrading 0.7.x -> 0.8.x so this may be something to watch out for.

Vintik:
Fixes #1525 

I'm going on vacation for 3 weeks so hopefully I didn't make any mistakes :)

### DIFF
--- a/components/data-table/__examples__/advanced.jsx
+++ b/components/data-table/__examples__/advanced.jsx
@@ -159,7 +159,11 @@ const Example = createReactClass({
 							property="accountName"
 							width="8rem"
 						/>
-						<DataTableColumn width="5rem" label="Close Date" property="closeDate" />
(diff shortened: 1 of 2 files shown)


interactivellama:
Thanks @Vintik. I hope you are having fun...umm, great timing ;-). 

I'm going to close this favor of #1532 which is the same fix--just from a different team.

Issue #1588, opened 2018-10-04
Input: Add styleInput prop 


seirykborys:
What do you think about adding such props as  errorText and inputStyle from datePicker to input?
I have already created PR with this changes, and it would be nice if you approve it.

interactivellama:
Datepicker takes an `<Input />` as `children`. I think that is what you are looking for. https://react.lightningdesignsystem.com/components/date-pickers/#prop-children

This allows any prop that is in `Input` to be passed in. I'm going to close, but feel free to re-open if you can't use `<Input errortext="test"/>`. https://react.lightningdesignsystem.com/components/inputs/#prop-errorText

Of even better, update the prop description to make it more clear for others!

interactivellama:
Re-opening. There is no style prop for Input right now. Please update the pull request to only add `styleInput` to Input. Please add a DOM markup test to confirm it works. By that I mean, add an example to storybook called `Custom Style` in https://github.com/salesforce/design-system-react/blob/master/components/input/__docs__/storybook-stories.jsx#L45 and it will automatically create a DOM snapshot test for your story.

seirykborys:
I have created new PR, thank's for the hints which helped me properly to do it.

PR #1590, opened 2018-10-04
Datepicker: Move custom input props to input render prop instead of children prop

### LINKED ISSUE #1524: Add 'required' Boolean Prop to DatePicker
After looking thru the docs and source code, it appears that the `<DatePicker>` component does not have a `required` boolean.

Mostly all other components have support for this - `<Input>`, `<TimePicker>`, `<Checkbox>`, `<Combobox>`, etc...

After reading thru the previous closed issues in this repo, it appears that `<SLDSSingleDatepicker>` [had this option at one point in time](https://github.com/salesforce/design-system-react/pull/242/files).

interactivellama:
Fixes #1524 and partial fix for #1588 in that it makes custom input props more obvious.

### Additional description

Console warning:
screen shot 2018-10-04 at 12 36 10 pm

---


components/date-picker/date-picker.jsx
```diff
@@ -103,6 +99,10 @@ const propTypes = {
 		today: PropTypes.string,
 		weekDays: PropTypes.array,
 	}),
+	/**
+	 * A [Dropdown](http://react.lightningdesignsystem.com/components/inputs/) component. The props from this `Input` component will be merged and override any default props. See [Component composition with prop spread] for more information on this methodology.
```
futuremint: I think you're missing a url here for "[Component composition with prop spread]".

interactivellama: Updated.

components/date-picker/date-picker.jsx
```diff
@@ -509,69 +560,32 @@ class Datepicker extends React.Component {
...
-				(this.props.children && this.props.children.props.value) ||
-				this.state.inputValue,
+		/**
+		 * 1. DEFAULT: Use default props or state if present.
+		 * 2. DEPRECATED API: Use old "first-level" props that have been deprecated.
+		 * 3. DEPRECATED API: If `children` is present, use props from single child which should be an `<Input/>`
+		 * 4. CURRENT API: Use composition with props spread merge from `input` prop.
+		 * */
+		const finalInputProps = {
+			...this.getDefaultInputProps({ assistiveText }),
```
futuremint: This is the only call to `getDefaultInputProps`.  And its also a fat arrow function so I don't see the point in passing in the `assistiveText` object as a parameter as opposed to just getting it inside of `getDefaultInputProps` itself.

components/date-picker/date-picker.jsx
```diff
@@ -509,69 +560,32 @@ class Datepicker extends React.Component {
...
+			...this.getTopLevelDeprecatedComponentProps({
```
futuremint: Same comment on this one.  I think maybe it might be cleaner to do all of this prop merging in one place.  Is this the best place to put this object creation?  Or should all of these methods just be inlined into one method called `getInputProps`?  Then the "logic" of getting the props for the input would be contained in one obvious method definition (albeit a long one) instead of spread out in 4 different methods across this file.  I think you can make a single method more readable with judicious use of clarifying variables inside of it rather than how this is currently written.

interactivellama:
What about moving the class methods into the render function? Do you like that better? Is it just proximity that's the issue? I agree that scrolling up to the `get` functions is a pain. I didn't realize that poor dev experience until `eslint` yelled at me.

futuremint:
I think there should just be _one_ class method that gets all the props needed for the input.  And inside that one method is all of the implementation details to do that (all of the defaults, overrides, combinations, etc.).  As it stands currently that implementation is partially inside the render function and partially outside (spread across the various "get" methods for the different props that are assembled inside the render function).  The render function should just be about rendering, so it should have one call to `getInputProps` (because the fact that they are assembled from different props is an implementation detail that will go away eventually and should be abstracted away).  Then the one `getInputProps` method has all the steps _currently_ required to assemble those props from the various sources, deprecated and otherwise.  As time progresses that will also contain the changes to how those props are derived to that one method potentially making future PR reviews a bit easier too.

interactivellama:
@futuremint Updated with "encapsulation of concerns" suggestion.

PR #1638, opened 2018-11-02
Object Switcher: Added props for combobox

### LINKED ISSUE #1600: Footer and header render in combobox
Component Lookup is deprecated and instead we have to use combobox. But I didn't find footer render in combobox. Can we add this functionality ?

seirykborys:
-  support cross-entity
- loading
- default value
- disabled
- Menu: Footer and Header

Fixes #1600 

### Additional description

---

* [x] The Accessibility panel of each Storybook story has 0 violations (aXe). Open [http://localhost:9001/](http://localhost:9001/).


seirykborys:
Hello Stephen, are you going to merge this PR or something wrong with it ? 
I saw that you deployed it month ago.

interactivellama:
@seirykborys I apologize if I was unclear.

The deployment was of your pull request branch's code and not a release of the library. You can see your code deployed to a Heroku (PaaS) "reivew app" here.
https://design-system-react-co-pr-1638.herokuapp.com/?selectedKind=SLDSCombobox&selectedStory=Inline%20Single%20Search%2FAdd%20Entities&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

My request was to add the header (Search salesforce) and the footer (Add item) of the menu items to the keyboard (arrow keys up/down) navigation to make the new menu items you added accessible to keyboard users. I was waiting for this feature before I continued to review and merge the code in.

The other option is that I believe I have time in the next few weeks to complete this ask if you'd prefer me to finish up the feature.

tanhengyeow:
@seirykborys @interactivellama Would be more than happy to move this forward if y'all are occupied :smile:

interactivellama:
@tanhengyeow Yes, that would be great. It's mostly done. I haven't looked at it in a while, but the primary issue is that the menu list header and footer are not in the index when the keyboard arrows are used. They should be accessible for keyboard users just like the other items.

Screen Shot 2019-03-12 at 11 51 31 PM

interactivellama:
@davidlygagnon Would you be willing to finish this PR up. This is mostly an accessibility ask to complete. You know Combobox so well, I think knocking this out wouldn't be too difficult.

davidlygagnon:
Sounds good - I'll wrap this up.

components/combobox/combobox.jsx
```diff
@@ -254,6 +254,40 @@ const propTypes = {
...
 	 */
 	variant: PropTypes.oneOf(['base', 'inline-listbox', 'readonly']),
+	/**
+	 * Default value of input. Provide uncontroled behaviour
+	 */
+	defaultValue: PropTypes.string,
+	/**
+	 * Provided to input to make it disabled
+	 */
+	disabled: PropTypes.bool,
```
davidlygagnon: Not sure why he added this? We already have `singleInputDisabled`. We would need to figure out what it means to disable the input for the cases where it's not a single select.

components/combobox/combobox.jsx
```diff
@@ -254,6 +254,40 @@ const propTypes = {
 	 * Changes styles of the input. Currently `entity` is not supported. _Tested with snapshot testing._
 	 */
 	variant: PropTypes.oneOf(['base', 'inline-listbox', 'readonly']),
+	/**
+	 * Default value of input. Provide uncontroled behaviour
+	 */
+	defaultValue: PropTypes.string,
```
davidlygagnon: Should we call this `defaultInputValue` ?

components/combobox/combobox.jsx
interactivellama: Please remove. There will never be an all variants disabled prop on a combobox. If folks want to hack with with `input` prop. They can I think.

components/combobox/combobox.jsx
```diff
@@ -254,6 +254,40 @@ const propTypes = {
...
+	 */
+	disabled: PropTypes.bool,
+	/**
+	 * Add loading spiner below the options
+	 */
+	loading: PropTypes.bool,
+	/**
+	 * Object for creating Add item below the options
+	 */
+	optionsAddItem: PropTypes.shape({
```
davidlygagnon: @interactivellama Wouldn't it be better to include this as part of the options object and have a different `type` attribute ? Something like `option.type='header'` and `option.type='footer'` Same question goes for `optionsSearchEntity` ? Under the hood, it was much cleaner to "merge" those 2 props with `options` to get the same keyboard nav behavior and highlighting on mouse over. Plus, we wouldn't need to have 2 extra props.

Other question is: If we keep as is, I'd suggest having a more generic name: `optionsHeaderItem`  and `optionsFooterItems` ?

Also, should it accept an array instead of just one prop?

components/combobox/combobox.jsx
interactivellama: Can this be added in the input prop?

components/combobox/combobox.jsx
interactivellama: I think the idea was that you wouldn't want to pass it through the filter like the other items on keypress--and that the data type is different since it's not data, it's UI chrome in a way.

As to type, I think I've seen multi-entity searching on Platform such as Accounts, Users, etc. so an array might make more sense. My original proposal called for arrays. I don't think I had caught it in the previous review. Good catch!

As to naming, I was hoping to keep it semantic just in case it ever changed places. such as Add Item moved to the top or something. Do you foresee different footer needs. As you say, it might be good to future proof it as a head/footer. I could go either way.

interactivellama:
Replacing with #1986

PR #1657, opened 2018-11-29
Tree: Update single select example to expand branch on click

### LINKED ISSUE #1656: Update example of expand on branch click to Tree
This example copies lightning:ui's single, item-only selection pattern and is not a part of SLDS, but is a common pattern on the SF platform. It seems like a good idea to have an example that does this for dev productivity.

interactivellama:
Fixes #1656

### DIFF
--- a/components/tree/__docs__/storybook-stories.jsx
+++ b/components/tree/__docs__/storybook-stories.jsx
@@ -39,7 +39,7 @@ storiesOf(TREE, module)
 		/>
 	))
 	.add('No Branch Select', () => (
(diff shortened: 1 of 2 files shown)


interactivellama:
@kwilloughby-sf 

Updating example on dev site:
https://design-system-react-co-pr-1657.herokuapp.com/?selectedKind=SLDSTree&selectedStory=No%20Branch%20Select&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

PR #1661, opened 2018-12-01
Input inlineHelpText can be `node` as well as `string`.

jgerigmeyer:
Matches type defined in https://github.com/salesforce/design-system-react/blob/master/components/input/index.jsx#L173.


interactivellama:
Thanks @jgerigmeyer !

PR #1714, opened 2019-01-11
Carousel React Component

### LINKED ISSUE #1697: Add Carousel Component
Would like to implement the \<Carousel\/\> component in React.
(https://www.lightningdesignsystem.com/components/carousel/)

Proposed props:
<pre>
Carousel.propTypes = {
	/**
	 * Array of objects with shape, needed for building a carousel item.
	 * A shape would have:
	 *   - id: 0   [REQUIRED]
	 *   - heading: "All my things! Part 0" // (Visible heading for default render)
	 *   - description: "Card 0 description."
	 *          // (Visible paragraph text for default render and navigation bullet text used in
	 *              assistive text tag)   [REQUIRED]
	 *   - imageAssistiveText: "First card accessible description."
	 *          // (image alt text, OPTIONAL, required if `src` is present),
	 *   - href: "https://www.salesforce.com" // (OPTIONAL, void(0) if not provided for default template
	 *   - src: "/images/examples/carousel-01.png"
	 */
	items: PropTypes.array.isRequired,
	/**
	 * Number of items to be displayed at a time in the carousel
	 */
	itemsPerPanel: PropTypes.number,
	/**
	 * Boolean showing whether the autoplay feature is available or not (by showing the autoplay 
         * button or not)
	 */
	hasAutoplay: PropTypes.bool,
	/**
	 * Interval for the autoplay iteration
	 */
	autoplayInterval: PropTypes.number,
	/**
	 * Boolean for displaying the navigation indicators left/right arrows) of the carousel
	 */
	hasPreviousNextPanelNavigation: PropTypes.bool,
	/**
	 * Boolean for infinite loop navigation
	 */
	isInfinite: PropTypes.bool,
	/**
	 * CSS classes that are applied to the component
	 */
	className: PropTypes.oneOfType([
		PropTypes.array,
		PropTypes.object,
		PropTypes.string,
	]),
	/**
	 * CSS style that is applied to carousel indicators
	 */
	indicatorStyles: PropTypes.object,
	/**
	 * Id of component, if desired. If not provided an id is automatically generated
	 */
	id: PropTypes.string,
        /**
	 * Accepts a custom carousel item rendering function
	 */
	onRenderItem: PropTypes.func,
	/**
	 * Description of the menu for screen-readers.
	 */
	assistiveText: PropTypes.shapeOf,
};

Carousel.defaultProps = {
	itemsPerPanel: 1,
	hasAutoplay: false,
	hasPreviousNextPanelNavigation: false,
	isInfinite: false,
        assistiveText: {
		nextPanel: 'Next Panel',
		previousPanel: 'Previous Panel',
	},
};
</pre>

PaulaIoana:
Fixes #1697

### Additional description

The <code>< Carousel /></code> component does not have a pre-set width or height. The component will take 100% from the parent container's width.
The number of items per panel can be customised through itemsPerPanel -> 1 (default), 3 or 5.
The <code>< Carousel /> </code> also has an auto-play mode. When rendered with hasAutoplay property set on True, the carousel items will auto iterate. The auto-play will stop, when either carousel reaches the last panel, or by interacting with the Pause button. When pressing Pause button, a continue auto-play functionality will be made available.


interactivellama:
I've set this up to auto-deploy the branch on push in case you want to send a link out to anyone.

https://design-system-react-co-pr-1714.herokuapp.com/?selectedKind=SLDSCarousel&selectedStory=Default%20With%201%20Item&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

djmccormick:
@interactivellama, have you had a chance to review?

interactivellama:
@djmccormick I will look at today. Thanks for pinging me. I thought it wasn't done because the tests are still failing.

It looks like maybe it's an old CSS file import.
https://travis-ci.org/salesforce/design-system-react/builds/480876147#L8263

components/carousel/__docs__/site-stories.js
```diff
@@ -0,0 +1,10 @@
+// This object is imported into the documentation site. An example for the documentation site should be part of the pull request for the component. The object key is the kabob case of the "URL folder". In the case of `http://localhost:8080/components/app-launcher/`, `app-launcher` is the `key`. The folder name is created by `components.component` value in `package.json`. The following uses webpack's raw-loader plugin to get "text files" that will be eval()'d by CodeMirror within the documentation site on page load.
+
+/* eslint-env node */
+/* eslint-disable global-require */
+
+const siteStories = [
+	require('raw-loader!@salesforce/design-system-react/components/carousel/__examples__/default.jsx'),
```
interactivellama: Unfortunately, the site example only has access to imports from `components/index.js` that means that you can't import the items file `carousel-items.js`. You can import that file in all the other storybook files though.

components/carousel/__examples__/carousel-items.jsx
```diff
@@ -0,0 +1,64 @@
...
+import Carousel from '~/components/carousel';
+import CarouselItem from '~/components/carousel/private/carousel-item';
+
+export const ITEMS = [
+	{
+		id: 1,
+		heading: 'Carousel Item 1',
+		description: 'Description for #1',
+		src: 'assets/images/carousel/carousel-01.jpg',
+		CTALabel: 'Get Started',
```
interactivellama: The `CTALabel` isn't a key we agreed upon. You can keep it in the items array, but only use it the custom `onRenderItem` callback.

components/carousel/__tests__/carousel.snapshot-test.jsx
```diff
@@ -0,0 +1,46 @@
...
+import Card from '~/components/card';
+import Carousel from '~/components/carousel';
+import { ITEMS } from '~/components/carousel/__examples__/carousel-items';
+
+import { renderMarkup } from '../../../tests/snapshot-helpers';
+
+import Default from '../__examples__/default';
+import '../../../styles/carousel/carousel.css';
+
+test('Carousel Component Default Snapshot - 1 tem, no navigation', () => {
```
interactivellama: typo: item

components/carousel/check-props.js
```diff
@@ -0,0 +1,28 @@
+/* Copyright (c) 2015-present, salesforce.com, inc. All rights reserved */
```
interactivellama: This is a new component. You shouldn't need this.

components/carousel/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+/**
+ * A carousel allows multiple pieces of featured content to occupy an allocated amount of space.
+ */
+class Carousel extends React.Component {
+	constructor(props) {
+		super(props);
+
+		const { items, itemsPerPanel } = this.props;
+
+		this.generatedId = shortid.generate();
```
interactivellama: Id generation should to occur `onComponentMount` and not in the constructor to make sure there aren't collisions.

components/carousel/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+		this.nrOfPanels = Math.ceil(items.length / itemsPerPanel);
+		this.stageItem = React.createRef();
+
+		this.state = {
+			translateX: -1000000,
+			currentPanel: 1,
+			isAutoPlayOn: this.props.hasAutoplay,
```
interactivellama: Props being loaded into state in the constructor is an anti-pattern, since this would only happen on component creation. I think we want to use the hasAutoplay props throughout the component in case it changed.

components/carousel/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+			isAutoPlayOn: !isAutoPlayOn,
+		});
+		actionToTake();
+	};
+	setTranslationAmount = (amount, cb) => {
+		this.setState(() => ({ translateX: amount }), cb);
+	};
+
+	setCurrentPanel = (amount, cb) => {
+		this.setState(() => ({ currentPanel: amount }), cb);
```
interactivellama: I like the use of the setState callback!

components/carousel/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+			[KEYS.RIGHT]: () => {
+				if (this.canGoToNext()) {
+					this.onNextPanelHandler();
+				}
+			},
+		};
+
+		if (keyDownCallbacks[event.keyCode]) {
+			EventUtil.trapImmediate(event);
+			keyDownCallbacks[event.keyCode](event.shiftKey ? 10 : 1);
```
interactivellama: What's going on here with the parameters? I'm not seeing them used in line 121 and 126.

components/carousel/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+					/>
+				</div>
+			</div>
+		);
+	}
+}
+
+Carousel.displayName = CAROUSEL;
+
+Carousel.propTypes = {
```
interactivellama: Please put these at the top of the file.

components/carousel/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+	/**
+	 * Boolean for displaying the navigation indicators (left/right arrows) of the carousel
+	 */
+	hasPreviousNextPanelNavigation: PropTypes.bool,
+	/**
+	 * Boolean for infinite loop navigation
+	 */
+	isInfinite: PropTypes.bool,
+	/**
+	 * CSS classes that are applied to the component
```
interactivellama: Please mention the node they are applied to--that is `.slds-carousel`. I'm not seeing `this.props.className` anywhere though

components/carousel/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+	/**
+	 * Id of component, if desired. If not provided an id is automatically generated
+	 */
+	id: PropTypes.string,
+	/**
+	 * Accepts a custom carousel item rendering function
+	 */
+	onRenderItem: PropTypes.func,
+	/**
+	 * Description of the carousel items for screen-readers.
```
interactivellama: Please list out keys in comments like this https://github.com/salesforce/design-system-react/blob/master/components/alert/index.jsx#L21

components/carousel/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+			</div>
+		);
+	}
+}
+
+Carousel.displayName = CAROUSEL;
+
+Carousel.propTypes = {
+	/**
+	 * Array of objects with shape, needed for building a carousel item.
```
interactivellama: Please alphabetize props to aid in finding one.

components/carousel/private/auto-play-button.jsx
```diff
@@ -0,0 +1,54 @@
...
+ *  AutoPlayButton is used to start/pause the autoplay iteration of the carousel
+ */
+const AutoPlayButton = (props) => {
+	const icon = props.isAutoPlayOn
+		? '/assets/icons/utility-sprite/svg/symbols.svg#pause'
+		: '/assets/icons/utility-sprite/svg/symbols.svg#play';
+
+	return (
+		<span className="slds-carousel__autoplay">
+			<button
```
interactivellama: The `ButtonIcon` component should be used here, other context globals like Icon paths will be used and will be hard-coded. See `import ButtonIcon from '../icon/button-icon';` 
```
<ButtonIcon
    category="utility"
    name="down"
    size="x-small"
/>
```
if we didn't want a button appearance. A standard button would work if we want a gray background.

components/carousel/private/carousel-indicators.jsx
```diff
@@ -0,0 +1,93 @@
...
+					'slds-is-active': isSelectedPanel,
+				}
+			);
+			return (
+				<li
+					className="slds-carousel__indicator slds-m-horizontal_xx-small"
+					role="presentation"
+					key={index}
+				>
+					<a
```
interactivellama: In order to move the focus to the active dot link. You will want to pass down from internal state an `indicatorHasFocus` and `isSelectedPanel` to the active dot. Then when those are true set the ref callback to call `focus()` on the node.

components/carousel/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+// Implements the [Carousel design pattern](https://www.lightningdesignsystem.com/components/carousel/) in React.
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+// ### shortid
+// [npmjs.com/package/shortid](https://www.npmjs.com/package/shortid)
+// shortid is a short, non-sequential, url-friendly, unique id generator
+import shortid from 'shortid';
+import '../../styles/carousel/carousel.css';
```
interactivellama: We can't import CSS into these components. I don't think one of the builds will take it (I think it's the ES6 build). We can document this in the component comments for the doc site though.

components/carousel/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+
+import checkProps from './check-props';
```
interactivellama: Not needed

package.json
```diff
@@ -162,6 +165,7 @@
 		"file-loader": "^0.11.0",
 		"find-remove": "^1.1.0",
 		"fs-extra": "^3.0.0",
+		"identity-obj-proxy": "^3.0.0",
```
interactivellama: I'm unsure about this. It seems only to be for development and won't be used in the NPM module

interactivellama:
This is very cleaning  written, logically ordered component and the sub-components are good places to encapsulate, too!

It is working well. Tabs are moving the the items. The assistive text is present. The auto-play works as expected.

There isn't a button in the SLDS example layout. The entire slide is the call to action. The default should reflect the SLDS pattern.

User experience feedback:

Tabbing to the navigation then using the arrows intermittently keeps the first dot blue and there two navigation dots are blue. This is occurring because the first dot that is tabbed to is keeping focus. On change of tab, focus should move to the active tab and the "old tab" should get a tabIndex of -1. The tree component has a good example of multiple nodes and only one focusable element.
screen shot 2019-01-24 at 12 46 29 am. Not moving the focus around also means that the assistive text does not get by the screenreader. 

The nav dots are not centered, but it looks like they are centered on the carousel container which as the set size from its parent `div`, so if the item got bigger it would be centered.
screen shot 2019-01-24 at 1 29 35 am

The [Platform Component](https://developer.salesforce.com/docs/component-library/bundle/lightning-carousel/example) is resizing with the width of the container. If the bottom or sides are cropped, is could hide content. I'm not sure a way around it. The navigation indicator variant is not resizing at all and the 1 item seems to be dependent on first load and does update with window resizing.
screen shot 2019-01-24 at 12 46 57 am

The first slide is cut off on the left and the inactive slide 2 is visible when the container (and windows) is smaller than the max-width specified (1280px). The multiple item examples get sized to the width on load. Are thinking that the consuming dev mix update the wrapping `div`'s width with an `onResize`? 

There does not appear to be an example with the "default" SLDS example. That this item has no button and the whole item receives focus and is clickable.

I seeing that tabbing through the items is making the carousel jump to hidden items. I'm not sure how to best fix this when there is a button in the item supplied by the consuming dev, but if the default was one clickable item, then you would have control to remove and add tabIndex to the only the active item. "When a panel becomes active, the `tabindex` of the other non-visible panels should be set to -1 while the active panel remains at 0." from the SLDS website.

components/carousel/index.jsx
PaulaIoana: done

components/carousel/index.jsx
PaulaIoana: done

tanhengyeow:
@PaulaIoana  I would be happy to help move forward this PR if you are occupied right now. It would also help me to learn from the work at the same time 😄

interactivellama:
Here is a screenshot of a three slide carousel working on the SF Platform tech stack for UX pattern reference (since the one one the SLDS website has only one slide/item:
carousel

I believe it is on the home page of a Platform dev org https://developer.salesforce.com/signup?d=70130000000td6N which is free to sign up and look around. The main difference from the current code is that the current code hides the overflow (sometimes you get 2.5 files when you resize the windows) while the carousel on the SF platform keep each slide to a 30% width. The carousel SF platform may not use SLDS markup/styles, but the UX pattern/behavior is correct.

interactivellama:
@tanhengyeow Thank you for offering. I'm going to assign this to @kevinparkerson The spec on the SLDS site doesn't help much with this component.

components/carousel/__examples__/carousel-items.jsx
kevinparkerson: going with `buttonLabel`, if present a button is rendered with that text, otherwise it is not rendered at all to adhere with the SLDS standard

components/carousel/__docs__/site-stories.js
kevinparkerson: @interactivellama need clarification on this one

components/carousel/index.jsx
kevinparkerson: @interactivellama need to discuss for clarification

interactivellama: I believe the line number is out of date. It's not https://github.com/salesforce/design-system-react/pull/1714/files#diff-b4ff71a0c207b08e9b6fcb834f32a53cR135

The question was: what is `event.shiftKey ? 10 : 1)` causing to occur on this line https://github.com/salesforce/design-system-react/pull/1714/files#diff-b4ff71a0c207b08e9b6fcb834f32a53cR136

components/carousel/__docs__/site-stories.js
interactivellama: The Codemirror-using demos on the doc site only have access to exports from `components/index.js`.

`carousel-items.js` is a file that gets imported into doc site stories. This will be undefined on the site. The items should be brought inline into the stories, so folks can copy and paste them out.

components/carousel/index.jsx
kevinparkerson: done

components/carousel/index.jsx
kevinparkerson: done

components/carousel/private/auto-play-button.jsx
kevinparkerson: done

package.json
kevinparkerson: deleted

components/carousel/index.jsx
kevinparkerson: @interactivellama need to discuss for clarification

components/carousel/private/carousel-indicators.jsx
kevinparkerson: done

components/carousel/index.jsx
kevinparkerson: done

components/carousel/index.jsx
kevinparkerson: @interactivellama need to discuss for clarification

components/carousel/__examples__/carousel-items.jsx
kevinparkerson: done

kevinparkerson:
This work was completed in #1918. This PR will be left open until that one is merged, then it will be closed. Thanks to everyone who helped contribute to this component, you did excellent work!

interactivellama:
@PaulaIoana @djmccormick I second that. 💯  The last thing we want to do is prevent you all from shipping product due to our component requirements. I will follow up with you when Carousel is released. That should be in the coming week.

PR #1754, opened 2019-02-19
Datepicker: Clarify MomentJS usage in props/docs

interactivellama:
### Additional description
Some questions were asked about MomentJS usage with Datepicker. This change tries to clarify the usage in the documentation.

---

### DIFF
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -18,6 +18,8 @@ newPRWelcomeComment: >
 
   Once feedback has been given, please reply to the feedback giver once the feedback on been addressed, so that they can continue the review.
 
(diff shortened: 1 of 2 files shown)


evanblack:
@interactivellama This definitely helps. Maybe add a note that you'll want to use both `formatter` and `parser` in conjunction for correct localization. I don't think that aspect was obvious to me. It looks like we can remove `formattedValue` in our situation as well.

interactivellama:
Great. `formattedValue` is just present for non-controlled folks (that is consumers not defining value--which is _highly discouraged_. It should be called `defaultFormattedValue`.

PR #1755, opened 2019-02-22
DataTable: Keyboard navigation

### LINKED ISSUE #1233: DataTable needs to be updated to latest SLDS accessibility spec
The current implementation of dataTable does not implement keyboard navigation of the grid. This would prevent a keyboard only user from accessing records in a large data set, without tabbing through every row.

See Inline Edit's Accessibility notes on ARIA Grid
https://www.lightningdesignsystem.com/components/data-tables/

garygong:
This implements table keyboard navigation as described in 
https://www.lightningdesignsystem.com/components/data-tables/.

To indicate that an element in the table is interactive, use the `InteractiveElement` Higher Order Component.  See `data-table/__examples__/interactive-elements.jsx` for an example.

```
const InteractiveButton = DataTableInteractiveElement(Button);
const InteractiveCheckBox = DataTableInteractiveElement(Checkbox);

const CustomDataTableCell = ({ id, children, ...props }) => {
	const cell = (
		<DataTableCell {...props}>
			<InteractiveCheckBox id={`${id}-checkbox`} labels={{ label: 'Option' }} />
			<InteractiveButton
				onClick={(event) => {
					event.preventDefault();
				}}
				label="Open"
			/>
		</DataTableCell>
	);
	return cell;
};
CustomDataTableCell.displayName = DataTableCell.displayName;
```

table nav

Fixes #1233


components/data-table/private/row.jsx
```diff
@@ -22,131 +23,156 @@ import {
...
-		 * Use this if you are creating an advanced table (selectable, sortable, or resizable rows)
-		 */
-		fixedLayout: PropTypes.bool,
-		id: PropTypes.string.isRequired,
-		item: PropTypes.object.isRequired,
-		onToggle: PropTypes.func,
-		rowActions: PropTypes.element,
-		selection: PropTypes.array,
-		tableId: PropTypes.string,
+const DataTableRow = (props) => {
```
garygong: I converted this to a function component so that the I could use the slightly more concise `useContext` syntax.

garygong:
@interactivellama When you get a chance, can you take a closer look at this PR?  I think I've covered all the scenarios now.

The only thing left to do is to write some browser tests for this.  I see some browser tests are failing, which I'll also look into.

Thanks!

interactivellama:
@garygong I got a question today about an inlineEdit pattern for DataTable from a Quip designer. Could one of your examples include a custom cell with a static input + edit button. I'm thinking that we could use the new `EditDialog` for this, too.

http://design-system-react-components.herokuapp.com/?selectedKind=SLDSPopover&selectedStory=Edit%20Dialog&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

interactivellama:
Here is code for the Inline Edit pattern with the non-accessible `InlineEdit` component 

https://gist.github.com/interactivellama/1bdef4c65f839aa3163a3f9c33088831

I'll explore doing this with the EditDialog Popover, send you my code, since InlineEdit component is deprecated.

ricochetdata:
@interactivellama When you get a chance, can you take a look at this commit: https://github.com/salesforce/design-system-react/pull/1755/commits/97bb29d558ca0019d473c44ae10c7d4c982a3966

Here's a summary of what's there:
- Add some basic browser tests
- Add required prop `keyboardNavigation`
- Add `checkProps` warning for using `false`

Here are the messages displayed in the console:
- Missing `keyboardNavigation` prop: 
```
Warning: Failed prop type: The prop `keyboardNavigation` is marked as required in `SLDSDataTable`, but its value is `undefined`.
```
- `keyboardNavigation` prop set to `false`:
```
Warning: [Design System React] The value of `false`, for prop `keyboardNavigation` will be removed in the next major version of SLDSDataTable. Please update your props. Use `true` instead, and ensure that interactive elements in the table are wrapped inside a `DataTableInteractiveElement`.
```
- `keyboardNavigation` prop set to `true`:  No warning message.

garygong:
@interactivellama When you get a chance, can you take a look at this commit: https://github.com/salesforce/design-system-react/pull/1755/commits/97bb29d558ca0019d473c44ae10c7d4c982a3966

Here's a summary of what's there:
- Add some basic browser tests
- Add required prop keyboardNavigation
- Add checkProps warning for using false

Here are the messages displayed in the console:
- Missing keyboardNavigation prop:
```
Warning: Failed prop type: The prop `keyboardNavigation` is marked as required in `SLDSDataTable`, but its value is `undefined`.
```
- keyboardNavigation prop set to false:
```
Warning: [Design System React] The value of `false`, for prop `keyboardNavigation` will be removed in the next major version of SLDSDataTable. Please update your props. Use `true` instead, and ensure that interactive elements in the table are wrapped inside a `DataTableInteractiveElement`.
```
- keyboardNavigation prop set to true: No warning message.

interactivellama:
I should have given you feedback sooner.

I'm not seeing this in the keyboard events:
* Arrow navigation still takes the user cell to cell in any direction, but focuses on the first actionable item in the cell, if there is one.

The focus is not moving with the arrow keys it's staying on the original tabbed into cell. Actually, I take that back. Sometimes it is sometimes it isn't, I can step through it sometime if you'd like. The up an down arrow causing the checkbox to take focus in the "selected" cells seems to be intermittent. Sometimes I have to press enter again to get into action mode.

Has any of this been tested with a Dropdown menu? I see that there is code to support it. Should se add one to test--for instance an action column?

Pressing the right arrow in the last column losing focus on the cell. It's possible I messed up the indexes when I merged in `upstream/master`.

interactivellama:
I went ahead and pushed up master merged in. That will update it to Storybook 5, too.

components/data-table/index.jsx
```diff
@@ -359,9 +425,141 @@ class DataTable extends React.Component {
...
+				},
+				activeElement,
+			}));
+		}
+	}
+
+	handleKeyDownRight() {
+		const newColumnIndex = Math.min(
+			this.state.activeCell.columnIndex + 1,
+			this.props.children.length
```
hajiz:
```suggestion
			this.props.children.length - (this.props.selectRows ? 0 : 1)
```

components/data-table/index.jsx
```diff
@@ -359,9 +425,141 @@ class DataTable extends React.Component {
...
+		}
+		if (!this.interactiveElements[rowIndex][columnIndex]) {
+			this.interactiveElements[rowIndex][columnIndex] = [];
+		}
+		this.interactiveElements[rowIndex][columnIndex].push(elementId);
+	}
+
 	// ### Render
 	render() {
 		const ariaProps = {};
```
hajiz: We can reset interactiveElements at render time to avoid it getting larger on every render (potential memory leak)

```suggestion
		this.interactiveElements = {};
		const ariaProps = {};
```

garygong:
@interactivellama If you have time, I'd appreciate your review on this PR.  I revived it from a couple years ago and fixed it up with the following changes:
- `keyboardNavigation` prop to enable `DataTable` keyboard navigation is optional. It is off by default, and existing consumers must opt in to the new behavior.
- Fixed bugs mentioned above (intermittent focusing, `Dropdown` support and test, losing focus)

Thanks a lot!

interactivellama:
Thank you for picking this up again. I will review this week.

garygong:
Thanks for reviewing!

I made small updates to change the mode back to navigation when focus leaves the table.  I also added a new `<InteractiveLink>` component that the `DataTable` will recognize as an interactive element; and also updated the Interactive Elements story to include links.

Check out the changes in this commit: https://github.com/salesforce/design-system-react/pull/1755/commits/423a5d36b60330e055ab954c8eaca6e33a34e14b

Thanks!

interactivellama:
Looks good to me. Dropdown snapshots and consumers of dropdowns need updating

interactivellama (approved):

PR #1765, opened 2019-02-26
Exposing autoComplete prop on SLDSInput and input prop on SLDSCombobox.

### LINKED ISSUE #1654: Expose input autocomplete attribute in Combobox and Input
Newer versions of Chrome completely ignore `autocomplete="off"`. So if you have something like a ComboBox of States and you want to allow a user to filter the options via input, Chrome might recognize the input as part of an Address and try to autofill the State.

A possible solution is to set `autocomplete="some_invalid_value"` which causes Chrome to not prompt for an autofill value.

The problem with that is, Firefox will default to using `autocomplete="on"` on invalid values.

Per https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion:

> In some cases, the browser will continue suggesting autocompletion values even if the autocomplete  attribute is set to off. This unexpected behavior can be quite puzzling for developers. The trick to   really enforcing non-autocompletion is to assign an invalid value to the attribute, for example:
> 
> `autocomplete="nope"`
> 
> Since this value is not a valid one for the autocomplete attribute, the browser has no way to match it,  and stops trying to autocomplete the field. Please note that this does not work in Firefox; it completely disregards the invalid value and reverts to default autocompletion behavior.

In my mind, the only solution is to do a quick browser check before setting any hard values for `autocomplete` like https://github.com/salesforce/design-system-react/blob/master/components/combobox/combobox.jsx#L809.

Thanks

davidlygagnon:
Fixes #1654 

### Additional description

- Adding autoComplete prop on SLDSInput component. 
- Adding input prop on SLDSCombobox component. This will allow users to override the inner-input props of the combobox with the ones passed into the input component.
- Added one more test for the new combobox input prop.

---


components/input/index.jsx
```diff
@@ -108,6 +108,10 @@ class Input extends React.Component {
 			label: PropTypes.string,
 			spinner: PropTypes.string,
 		}),
+		/**
+		 * Disabled brower's autocomplete when "off" is used.
+		 */
+		autoComplete: PropTypes.string,
```
davidlygagnon: Should I add a new test just for this prop ?

components/combobox/combobox.jsx
```diff
@@ -139,6 +139,11 @@ const propTypes = {
 	 * HTML id for component. _Tested with snapshot testing._
 	 */
 	id: PropTypes.string,
+	/**
+	 * An [Input](https://react.lightningdesignsystem.com/components/inputs) component.
+	 * The props from this component will override any default props.
+	 */
+	input: PropTypes.node,
```
davidlygagnon: Now allowing to pass an input as a prop.

components/combobox/combobox.jsx
```diff
@@ -1026,6 +1037,7 @@ class Combobox extends React.Component {
 									props.value
 								: props.value
 						}
+						{...userDefinedProps.input}
```
davidlygagnon: Input props override any props set above this line.

components/input/index.jsx
interactivellama: All props that can be changed either need to be tracked via snapshot tests or mocha tests.

interactivellama: I think autocomplete='test' is in one of the snap shots.

interactivellama: Oh, I'm sorry. I misread this. I thought it was in combobox. Yes, a snapshot is fine.

PR #1772, opened 2019-02-28
Correct Few Typos in Contributor's Guidelines

aswinshenoy:
I was going through the contributing guideline, as I was planning to implement the progress bar component from LDS.

Fixes #
*  Markdown hyperlink syntax error in the part where it mentions `Create dev storybook stories`, there was a missing ```[```, which I have added.
*  Path of ```constants.js``` is mentioned to be ```/components/constants.js``` but actually is in ```/utitlites/constants.js```


interactivellama:
@aswinshenoy Thank you for the corrections!

PR #1777, opened 2019-03-01
Replace dismissible with disableClose in Modal

### LINKED ISSUE #503: Replace Modal's default true prop: dismissable
JSX props should not be true by default.

My suggestion is to deprecate `dismissable` and remove it from the propTypes and use the following.

```
/**
 * If true, Modals cannot be dismissed by clicking on the close icon or pressing esc key.
 */
disableClose: React.PropTypes.bool,
``
```

This should first be warned with a console error with dismissable is used.

This should not be a breaking change.

aswinshenoy:
Fixes #503 

### Additional description

Replaced instances of the prop-type ```dismissible``` with ```disableClose``` 

---


components/modal/index.jsx
```diff
@@ -79,11 +79,11 @@ const propTypes = {
 	 */
 	directional: PropTypes.bool,
 	/**
-	 * If true, Modals can be dismissed by clicking on the close icon or pressing esc key.
+	 * If true, Modals cannot be dismissed by clicking on the close icon or pressing esc key.
 	 */
-	dismissible: PropTypes.bool,
```
davidlygagnon: Removing this will break consumers of `dismissible`. We need to deprecate `dismissible` and the new prop `disableClose`. Take a look at `check-props.js` to see how it's done.

components/modal/index.jsx
```diff
@@ -79,11 +79,11 @@ const propTypes = {
...
+	disableClose: PropTypes.bool,
 	/**
-	 * If true, Modals can be dismissed by clicking outside of modal. If unspecified, defaults to dismissible.
+	 * If true, Modals can be dismissed by clicking outside of modal. If unspecified, defaults to disableClose.
 	 */
 	dismissOnClickOutside: PropTypes.bool,
```
davidlygagnon: We have another prop called `dismiss...`. I would think we would want to do the same to be consistent?

components/modal/index.jsx
aswinshenoy: I will patch this by evening :)

components/modal/index.jsx
aswinshenoy: hey @davidlygagnon, I didn't get you on this? can you kindly clarify :)

components/modal/index.jsx
aswinshenoy: I am adding the following into `check-props.js` to show up a message -
```
deprecatedProperty(
			COMPONENT,
			props.dismissible,
			'dismissible',
			"disableClose",
			'The prop `dismissible` has been deprecated, please use `disableClose` instead.'
		);
```
@davidlygagnon would that be enough? Thank you :)

components/modal/index.jsx
tanhengyeow: I think what @davidlygagnon means is that we keep the semantics similar, e.g. change all `dismisss...` to `disable...`, similar to the change from `dismissable` to `disableClose`.

interactivellama: @davidlygagnon What name would you recommend?

davidlygagnon: Shouldn't that follow the same pattern as `disableClose`?
```
/**
 * If true, Modals can be dismissed by clicking outside of modal. If unspecified, defaults to disableClose.
 */
disableCloseOnClickOutside: PropTypes.bool,
```

interactivellama: Got it. Yes, great suggestion.

`dismissOnClickOutside` is a default true also, it's just not in the default props, since it defers to the `dismissible` if undefined.

`dismissOnClickOutside` should become `disableCloseOnClickOutside` also.

interactivellama:
@davidlygagnon I've created #1916 to handle other default true prop.

davidlygagnon:
Cool thanks!

PR #1794, opened 2019-03-05
Update README.md to run install command in right directory while setting up the code base locally

pradumangoyal:
Fixes #

* In setting up section in the `README.md`  file, `npm install` command was run in the same directory whereas it is required to open the cloned directory for which I have added the `cd design-system-react` command


interactivellama:
Thanks you @pradumangoyal for the correction!

Issue #1708, opened 2019-01-10
DataTable Radio button with fixedLayout and fixedHeader is broken


gzhang86:
With 0.9.3, when creating a single selected radio button DataTable with fixed header:
<DataTable
	fixedLayout
	fixedHeader
	items={...}
	selectRows='radio'
	onRowChange={...}
	selection={...}
>

We are hitting the following in DataTable. resizeFixedHeaders():
```
              var cellFixed = column.querySelector('.slds-cell-fixed');
              var linkFixed = cellFixed.firstChild;

              if (cellFixed) {
                cellFixed.style.left = "".concat(columnLeft - wrapperLeft, "px");
                **linkFixed**.style.width = "".concat(column.offsetWidth, "px");    // linkFixed is null
              }
```

interactivellama:
if (cellFixed) {
    cellFixed.style.left = `${columnLeft - wrapperLeft}px`;
}

if (linkFixed) {
    linkFixed.style.width = `${column.offsetWidth}px`;
}
```

davidlygagnon:
Thanks @interactivellama. I asked him to file this bug because it's preventing us from using > 0.9.2. Let me know if you need me to create the PR for this fix.

interactivellama:
I made one. #1710 Please review.

PR #1802, opened 2019-03-06
Data table fixed header scrolling bugfix

### LINKED ISSUE #1696: DataTable: Scroll/render bug on tab
Tabbing between the headers creates this bug when the table is hidden. It happens in Chrome and FireFox. Tabbing between items in the actual table data does not create this bug

screen shot 2018-12-20 at 5 09 46 pm

kevinparkerson:
Fixes #1696

This was an interesting bug! When tabbing and therefore focusing on headers outside the visible area the browser has to make a decision on how to bring them into view, which apparently defaults to "scroll the parent container." Since the headers are absolutely positioned as far as the rendering engine is concerned that parent is `.slds-table_header-fixed_container` rather than the table header that actually contains them. The browser then "scrolls" that parent left by an arbitrary amount in order to bring the focused element into view. Since `.slds-table_header-fixed_container` actually has `overflow: hidden` no scrollbar is shown, and `.slds-table_header-fixed_scroller` appears "clipped" because it is retaining the actual width of the parent container as it is scrolled left to reveal the hidden content. The effect looks quite goofy but it actually makes sense once you understand what's going on!

I tried a few different approaches to fixing this bug such as making the container have `overflow-x: auto` and the scroller `overflow-y: auto`, but that introduced other visual bugs. The simplest, most elegant solution I found was to add an `onScroll` event to the container that sets its scrollLeft value to a variable and checks if it surpasses 0. If so, the container's scrollLeft is reset to 0. Since both the container and scroller share the same width, the pre-reset scrollLeft value can then be applied to the scroller itself and effectively bring the focused content into view. The scroller's listeners then kick in and make sure to reposition / resize the headers correctly as expected, and everything looks as it should!

---


components/data-table/index.jsx
```diff
@@ -316,14 +316,10 @@ class DataTable extends React.Component {
 					}
 
 					const cellFixed = column.querySelector('.slds-cell-fixed');
-					const linkFixed = cellFixed.firstChild;
 
 					if (cellFixed) {
 						cellFixed.style.left = `${columnLeft - wrapperLeft}px`;
-					}
-
-					if (linkFixed) {
```
kevinparkerson: @interactivellama why was this altered? I had styles to make headers fill the cellFixed element provided it was properly sized. If there was something I missed I would prefer to add `width: 100%` to whatever it was and only resize the cellFixed element rather than attempt to size it's immediate child manually in this way

components/data-table/__examples__/fixed-header.jsx
```diff
@@ -204,9 +204,10 @@ class Example extends React.Component {
 			<div
 				style={{
 					height: '200px',
-					width: '100%',
+					width: '800px',
 				}}
 			>
+				<style dangerouslySetInnerHTML={{ __html: '.dsr_data-table-fixed-header-scroll-test { width: 1600px; }' }} />
```
kevinparkerson: @interactivellama this was added for testing... do we want to leave this stuff in there and make a vertical / horizontal fixed header story example the default that shows up on the site? Or should I make another example with horizontal scrolling in order to preserve the simpler vertical-only scrolling fixed header example?

interactivellama: I'd lean toward another story, although if we can just pass in a prop that modifies the same story, that'd probably be best.

components/data-table/index.jsx
interactivellama: I believe the value was null in the radio group variant which resulted in a key call on an undefined object.

https://github.com/salesforce/design-system-react/issues/1708

components/data-table/index.jsx
```diff
@@ -517,6 +513,16 @@ class DataTable extends React.Component {
 				<div
 					className="slds-table_header-fixed_container"
 					style={{ height: '100%' }}
+					onScroll={(e) => {
```
interactivellama: Can we add more conditions to this? Maybe something like if horizontal scrolling can occur (that is something like measuring `clientWidth`?).

components/data-table/index.jsx
kevinparkerson: Resolved via meeting

components/data-table/__examples__/fixed-header.jsx
kevinparkerson: @kevinparkerson: create a `style` prop that gets applied to the table. Create a separate fixed header horizontal story that demonstrates horizontal scrolling without adding it to doc site. Update doc site to refer to using `style` prop to obtain horizontal scrolling (have a note in `style` prop description and a note in the top description of the component referencing the `style` prop)

kevinparkerson:
@interactivellama this is ready for re-review

Issue #1840, opened 2019-03-13
Button Overlap in Alert Example in the Storybook


sharaththegeek:
For SLDSAlert -> Close Alert in the Storybook, the "Toggle Alert" Button is overlapped when the alert appears. This makes it look clumsy and intrusive as the same button also serves the function of dismissing the alert. 

While this is not a performance issue or not something that would affect practical usage while using the Lightning Design System, I believe that since this is a design project, it is necessary to reflect our core commitment of clean user interfaces, no matter how insignificant the location is. :) 

<img width="1436" alt="Screenshot 2019-03-13 at 8 14 46 PM" src="https://user-images.githubusercontent.com/33923577/54287959-a8d9c600-45cc-11e9-8f33-116318ac3390.png">

sharaththegeek:
Need some help with #1841 
Can someone help me identify why Travis CI failed. I am not able to figure it out.

abhinavsagar:
@sharaththegeek You are overwriting `dev:true` at multiple places that's why travis CI is not able to recognize it.

sharaththegeek:
@abhinavsagar I removed the package-lock.json file from the fork and replaced it with the clean default. It still fails. 

I have no idea why dev:true was overwritten in the first place. Any opinions on how it occurred?

PR #1841, opened 2019-03-13
Corrected SLDS Close Alert Button placement

sharaththegeek:
Corrected the placement of the "Toggle Alert" button in SLDSAlert -> Close Alert so that it is not overlapped by the alert when it appears.

### DIFF
--- a/components/alert/__examples__/close-alert.jsx
+++ b/components/alert/__examples__/close-alert.jsx
@@ -19,6 +19,7 @@ class Example extends React.Component {
 			<IconSettings iconPath="/assets/icons">
 				<div>
 					<Button
+					    style={{ marginTop: 25 }}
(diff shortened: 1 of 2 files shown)


sharaththegeek:
Corrects Issue #1840

interactivellama:
@sharaththegeek TravisCi is saying there are lint issues.

components/alert/__examples__/close-alert.jsx
```diff
@@ -19,6 +19,7 @@ class Example extends React.Component {
 			<IconSettings iconPath="/assets/icons">
 				<div>
 					<Button
+					    style={{ marginTop: 25 }}
```
interactivellama: All CSS should have units on them such as `px`.

PR #1867, opened 2019-03-24
Resolve console warns and errors

### LINKED ISSUE #1849: Remove console warnings & errors
While running `npm test` one get following console warnings and errors which are maybe due to not updating Examples while updating props of elements. This issue can be resolved by updating examples as required.

console-errors-test

pradumangoyal:
Fixes #1849 

@interactivellama @futuremint, Can you review it and guide me through some problems I am facing while resolving the issue.
 
* [x] Resolve PopoverTooltip warning which expects the element to be tabbable and id in Filter example needed in type `string`
* [ ] Resolve warning due to use of `offset`

### Additional description

The problem was not only with the examples but with the components that were using those components in it. Other components(here Combobox) were not accordingly updated when changes were made in the props of PopoverTooltip.

Now upper two errors don't appear while running `npm test` as shown following

image

`offset` warning is due to the following code [here](https://github.com/salesforce/design-system-react/blob/44cc2624cd0edb0d88408e514a544541ddb92a95/components/filter/index.jsx#L170). Is there any alternative to produce the same result?

image
image

While, resolving this I came across the following console warning in an example that is corrected in the change.

image

Also, should I update the tests according to these changes also?

---

### DIFF
--- a/components/combobox/private/menu.jsx
+++ b/components/combobox/private/menu.jsx
@@ -2,6 +2,7 @@
 /* Licensed under BSD 3-Clause - see LICENSE.txt or git.io/sfdc-license */
 
 /* eslint-disable jsx-a11y/interactive-supports-focus */
+/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
 
 import React from 'react';
 import PropTypes from 'prop-types';
@@ -189,6 +190,7 @@ const Menu = (props) => {
 								}
 					}
 					role="option"
+					tabIndex="0"
 				>
 					{optionData.icon && !props.menuItem ? (
 						<span className="slds-media__figure">{optionData.icon}</span>
--- a/components/menu-picklist/__examples__/tooltip-list-item.jsx
+++ b/components/menu-picklist/__examples__/tooltip-list-item.jsx
@@ -1,3 +1,5 @@
+/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
+
 import React from 'react';
 
 import IconSettings from '~/components/icon-settings';
@@ -10,7 +12,9 @@ const ListItemRenderer = (props) => (
 		align="bottom left"
 		content={`${props.label} tooltip on bottom left`}
 	>
-		<p className="slds-truncate">{props.label} (Hover for tooltip)</p>
+		<p className="slds-truncate" tabIndex="0">
+			{props.label} (Hover for tooltip)
+		</p>
 	</Tooltip>
 );
 
--- a/components/filter/__examples__/default.jsx
+++ b/components/filter/__examples__/default.jsx
@@ -9,8 +9,8 @@ import Combobox from '~/components/combobox';
 
 const options = {
 	'show-me': [
-		{ id: 1, label: 'All Products', value: 'all-products' },
-		{ id: 2, label: 'All Wackamoles', value: 'all-Wackamoles' },
+		{ id: '1', label: 'All Products', value: 'all-products' },
+		{ id: '2', label: 'All Wackamoles', value: 'all-Wackamoles' },
 	],
 };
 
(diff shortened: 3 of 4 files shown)


components/menu-picklist/__examples__/tooltip-list-item.jsx
```diff
@@ -10,7 +12,9 @@ const ListItemRenderer = (props) => (
 		align="bottom left"
 		content={`${props.label} tooltip on bottom left`}
 	>
-		<p className="slds-truncate">{props.label} (Hover for tooltip)</p>
+		<p className="slds-truncate" tabIndex="0">
```
interactivellama: `jsx-a11y/no-noninteractive-tabindex` is triggering because it doesn't have an ARIA role such as `button`. `role="button"` should fix this.

components/combobox/private/menu.jsx
```diff
@@ -189,6 +190,7 @@ const Menu = (props) => {
 								}
 					}
 					role="option"
+					tabIndex="0"
```
interactivellama: The warning is primarily present so that engineers do not attach tooltips to appear on mouse hover over `div` tags or something like that. Combobox is accessible for screen readers the way it is right now.

Menu listbox item menus shouldn't have `tabIndex`. This will result in focus being changed to them when clicked with the mouse. I think the best direction would be modifying `ToolTip` to not trigger the warning.

Maybe an explicit override prop like`IgnoreIsTriggerTabbable` and then add that to the tooltip usage in the Combobox? What do you think?

interactivellama:
I'm not sure about the offset usage. It's there because it's a reminder that the code isn't "implemented correctly"--however I can see how most contributors would believe it might be their fault to have an error in their tests.

@davidlygagnon I think you all have Filters on the left side of the screen in a panel. Do you have a recommendation? Maybe supporting right aligned filters would be a good dependency on the SLDS project? 

This is the line of code that is causing the error in the test suite: https://github.com/salesforce/design-system-react/blob/44cc2624cd0edb0d88408e514a544541ddb92a95/components/filter/index.jsx#L170

components/menu-picklist/__examples__/tooltip-list-item.jsx
pradumangoyal: That's not allowed to assign `role` to non-interactive, showing following error.
image

Similar approach is also used in other parts also like `components/app-launcher/tiles.jsx`.

components/combobox/private/menu.jsx
interactivellama: You are approved to add `IgnoreIsTriggerTabbable` to Tooltip and implment that solution in Combobox.

davidlygagnon:
@interactivellama Yeah, I even wonder if we need this hardcoded `-35px` offset? Seems like it's used in the calculation of `left` when passed to popper.js : 

https://github.com/salesforce/design-system-react/blob/44cc2624cd0edb0d88408e514a544541ddb92a95/components/utilities/dialog/index.jsx#L265-L268

This was added a while back here but not sure why the `-35px`: https://github.com/salesforce/design-system-react/commit/f27a178c8bc67f4839e25f344dc1d91bd87918a5

It seems we consume SLDSFilter component with `align=right` in a few places in the codebase.

What do you mean by : "Maybe supporting right aligned filters would be a good dependency on the SLDS project?" ?

interactivellama:
@davidlygagnon I was thinking that we might be able to consume additional CSS from SLDS and the offset could be removed. 

At this point, I don't think it will be prioritized by the SLDS team. I'm going to push up code without the offset which will move the popover 35px farther in to the Filter "button", but I don't believe anyone will notice.

The filter panel is looking okay without it from what I can tell.

Screen Shot 2019-04-17 at 12 23 20 AM

davidlygagnon:
Just to circle back on this - I think that's fine.

PR #1902, opened 2019-04-03
Breadcrumb: Ability to add overflow menu

### LINKED ISSUE #1898: Breadcrumbs: Ability to add overflow menu
Mentioned in #1500 by @interactivellama, ability to add overflow menu to the breadcrumbs is currently missing in DSR while it is present in SLDS.

I would like to work on this issue and will submit a PR soon.

pulkonet:
There was no option to add overflow menu in the current version
of breadcrumb. This commit aims at adding that functionality by
adding a new prop `parentIndex` which slices the trail into a
menu trail and a regular trail, which can be then used to render
overflow menu and regular breadcrumb. The prop can have a better
name though, as it might not be very obvious in the first glance.

Fixes #1898 

### Additional description

---


interactivellama:
This is to the spec of the SLDS Breadcrumb Blueprint. Menus should be accessible and allow keyboard navigation, and close when clicked off of, etc.

Can you implement the menu using the `Dropdown` component?

pulkonet:
@interactivellama I have updated the code to use Dropdown instead of the earlier blueprint I was using, but there is one issue: Dropdown needs to have a unique ID for supporting keyboard navigation and ARIA, and that is uniquely generated by this
```
this.generatedId = shortid.generate();
```
This is rendering the Jest snapshot to be different and the Travis CI test is failing. What can be done here?

interactivellama:
Please add an `id` prop to Breadcrumb and and then concatenate a string something like `${props.id}-dropdown` that will pass id into dropdown, so it can be controlled by breadcrumb users--then use the id in your example code in order to keep the same id for every time the test is run.

Here is an example of Colorpicker passing id down to Tabs subcomponent: https://github.com/salesforce/design-system-react/blob/master/components/color-picker/index.jsx#L247

https://github.com/salesforce/design-system-react/blob/master/components/color-picker/index.jsx#L357

pulkonet:
I can do that, but I thought we want to generate the id prop to be unique using the above method. Will revert soon.

interactivellama:
Here is a blog post from a non-Salesforce employee about the pattern. http://bradfrost.com/blog/post/enforcing-accessibility-best-practices-with-automatically-generated-ids/

In a perfect world, the consuming dev would write their own unique id for the page, but that doesn't always happen so `shortid` is a fallback that guarantees it's accessible.

pulkonet:
Went through the blog and cleared the doubt, thanks for it. I have added the `id` prop to Breadcrumb and the PR can be reviewed now.

interactivellama:
@garygong @davidlygagnon Did you want me to review this pull request or did you want me to?

davidlygagnon:
@interactivellama I will review it. Thanks.

components/breadcrumb/index.jsx
```diff
@@ -33,18 +35,32 @@ const defaultProps = {
...
+		menuTrail = menuTrail.map((item) => ({
+			label: item.props.children,
+			value: item.props.children,
+		}));
+		trail = trail.slice(parentIndex);
+	}
 	return (
 		<nav role="navigation" aria-label={assistiveText}>
 			<ol className="slds-breadcrumb slds-list_horizontal">
+				{menuTrail.length > 0 && (
```
davidlygagnon: Could we instead expose a `BreadcrumbDropdown` instead of `parentIndex` that the consumer populates for the overflow menu? That way, it would simplify the code a lot (no need of splitting `trail`) and users won't need to keep track of an index where the cut should happen for overflow items? What do you think?

components/breadcrumb/private/menu.jsx
```diff
@@ -0,0 +1,39 @@
...
+const defaultProps = {};
+const Menu = (props) => (
+	<Dropdown
+		id={`${props.id}-dropdown`}
+		assistiveText={{ icon: 'Show More' }}
+		threedots
+		iconCategory="utility"
+		iconName="threedots"
+		iconVariant="bare"
+		onSelect={props.onSelect}
```
davidlygagnon: It doesn't seem like we can pass this from `breadcrumb`?

components/breadcrumb/index.jsx
pulkonet: That can be done too. I will update the code accordingly and expose the `BreadcrumbDropdown` subcomponent.

interactivellama:
What's the update on this? @davidlygagnon

components/breadcrumb/__docs__/storybook-stories.jsx
```diff
@@ -4,10 +4,12 @@ import { storiesOf } from '@storybook/react';
 import { BREADCRUMB } from '../../../utilities/constants';
 import Base from '../__examples__/base';
 import OneItem from '../__examples__/one-item';
+import OverFlowMenu from '../__examples__/base-with-overflow-menu';
```
davidlygagnon: Minor: `import OverflowMenu`

components/breadcrumb/__examples__/base-with-overflow-menu.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+
+import IconSettings from '~/components/icon-settings';
+import BreadCrumb from '~/components/breadcrumb'; // `~` is replaced with design-system-react at runtime
```
davidlygagnon: minor : `import Breadcrumb`

components/breadcrumb/index.jsx
```diff
@@ -20,20 +20,56 @@ import componentDoc from './docs.json';
...
+	/**
+	 * A unique ID is needed in order to support keyboard navigation, ARIA support, and connect the dropdown to the triggering button.
+	 */
+	id: PropTypes.string,
+	/**
+	 * Overflow menu of the type [Dropdown](/components/menu-dropdowns)
+	 */
+	overflowDropdownMenu: PropTypes.node,
+	/**
+	 * An array of react elements presumably anchor elements.
```
davidlygagnon: Maybe we could say: "An array of anchor elements that define the path to the current record."

components/breadcrumb/index.jsx
```diff
@@ -20,20 +20,56 @@ import componentDoc from './docs.json';
...
+	 */
+	trail: PropTypes.array,
```
davidlygagnon: We should make this required since the breadcrumb component would be useless without this.

davidlygagnon:
It looks good - just a few items to cleanup in the PR. 

@interactivellama One though: the `trail` props has been around before this PR and suggest the consumer to create "anchor tags". It's hard to enforce. I would think it might be better to have a <BreadcrumbTrail> component specifying a label, href at a minimum. That way we control rendering of the `<a>` tags since I believe trail items should not be anything else than that.

What do you think?

components/breadcrumb/index.jsx
pulkonet: Yes, sure.

davidlygagnon:
@pulkonet Thanks, looks good. I will merge this shortly.

interactivellama:
@davidlygagnon I agree with your comment on having an array of collections, probably with an `onClick` that returns an `id`. At this point, it would be a low priority, due to the simplicity of the component and other items in the backlog.

PR #1939, opened 2019-04-22
ButtonGroup: Adds `list` variant

### LINKED ISSUE #1371: Update Button Group to use list variant
This should allow tooltips trigger `div`'s to be wrapped around buttons within button group (such as in a page header and prevent the following. See list example in  https://www.lightningdesignsystem.com/components/button-groups/

27f603fc-e3d4-11e6-8270-d72288554c08

pulkonet:
Fixes #1371 

Also adds `list` variant for Tooltip component to work well with ButtonGroup.

Screenshot 2019-04-23 at 2 04 58 AM

### Additional description

---


components/button-group/index.jsx
```diff
@@ -102,6 +102,16 @@ const ButtonGroup = (props) => {
 			</fieldset>
 		);
 	}
+	if (props.variant === 'list') {
+		return (
+			<ul
+				className={classNames('slds-button-group-list', props.className)}
+				role="group"
```
interactivellama: Group is implicit here and not needed--unlike in base variant

components/tooltip/__examples__/button-group.jsx
```diff
@@ -12,35 +12,44 @@ class Example extends React.Component {
...
-						iconName="down"
-						iconVariant="border-filled"
-						onSelect={(item) => console.log('selected', item)}
-						options={[
-							{ label: 'A Option', value: 'A0' },
-							{ label: 'B Option', value: 'B0' },
-							{ label: 'C Option', value: 'C0' },
-						]}
-					/>
+					<Tooltip
```
interactivellama: Tooltip isn't working for me here. You are welcome to just remove it if you want. I'm guessing this is a side affect of Dropdown.

pulkonet: Removing the tooltip is adding double borders ( no overlap ) and is also not adding the `corner radius` to dropdown button.

pulkonet: I am wrapping it in an `li` tag for now.

components/button-group/index.jsx
pulkonet: Removed.

components/tooltip/__docs__/__snapshots__/storybook-stories.storyshot
```diff
@@ -1992,41 +1991,45 @@ exports[`DOM snapshots SLDSPopoverTooltip Button Group 1`] = `
...
-      className="slds-dropdown-trigger slds-dropdown-trigger_click slds-button_last"
-      id="options"
-      onClick={[Function]}
-      onFocus={null}
-      onKeyDown={[Function]}
-      onMouseEnter={null}
-      onMouseLeave={null}
+    </li>
+    <li
+      triggerClassName="slds-button_last"
```
interactivellama: I'm getting this warning on the example added at http://localhost:9001/?selectedKind=SLDSPopoverTooltip&selectedStory=Button%20Group&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

Screen Shot 2019-05-14 at 1 56 03 AM

The ButtonGroup is expecting `Button` child components which have a `triggerClassName` but, but `li` element tags do not. Please fix this warning.

pulkonet:
```if (zeroIndexLength > 0) {
		children = React.Children.map(props.children, (child, index) => {
			let newChild;
			if (index === zeroIndexLength) {
				newChild = React.cloneElement(child, {
					triggerClassName: 'slds-button_last',
				});
			}

			return newChild || child;
		});
	}
```
The error is being generated from the above LOC (present in `button-group`). Adding a type check to prevent the prop `triggerClassName` being added to `li` element will resolve the warning. Shall I go ahead with that ? @interactivellama

interactivellama: Yes. I wouldn't recommend variable type checking though. Since `triggerClassName` is an implementation details of `SLDSButton`, you could check for that and only apply the prop if it's an instance of a `SLDSButton`.

pulkonet: But we are adding the prop to the instance of `Tooltip` rather than `SLDSButton`. Because the children of the `ButtonGroup` are `Tooltip`s and not `Button`s

pulkonet: Do you have any other suggestions in mind?

interactivellama: You are exactly right. My mistake. The prop will only work on a `Tooltip` component. So we want to test if the child is a `Tooltip`. 

Here is a displayName test in another component: https://github.com/salesforce/design-system-react/blob/master/components/builder-header/nav.jsx#L33

pulkonet: I have added the `displayName` test for the `Tooltip`

components/tooltip/__examples__/button-group.jsx
pulkonet: The error is being generated from here, we can't be adding props to DOM elements in `camelCase`, they need to be in `lowercase`.

So, it would be best if we can either type check the element and filter out `li` or let `Tooltip` be here unless it is creating problems.

@interactivellama

components/tooltip/__docs__/__snapshots__/storybook-stories.storyshot
pulkonet: It is creating issues elsewhere in the projects wherever `ButtonGroup` is being used. My proposed solution would be to replace `li` with `Tooltip` and let it be there.

interactivellama: If you are referring to PageHeader, https://travis-ci.org/salesforce/design-system-react/builds/534113631#L7524 for instance.

Like code, Jest snapshot markup tests are not fixed in stone. Similar to version control such as "git", the snapshots are a way to track history over time (humans are very bad at this!) and know what has changed across time. If there is a bug or a better way to do something, updating the Jest DOM snapshots is a perfectly fine thing to do. 

In summary, it's OK to change the DOM snapshots, but only if it's a better solution. They are there so that you know what is changing and can decide if you really want to update the markup. Please review the changes across the library per the "failed" Jest tests visually and decide if updating the markup is best and report back.

components/tooltip/__examples__/button-group.jsx
interactivellama: A tooltip around a Dropdown is fine. My issue is that the Tooltip wasn't being triggered on mouseover / "not working". I was thinking that if test the displayName === Tooltip then the extra props wouldn't get passed unless Tooltip was present. I might be misunderstanding though.

pulkonet: I am reverting to a tooltip around the dropdown instead of li. I will also open another issue for the Tooltip not being triggered and we can work on it there.

kevinparkerson:
@davidlygagnon I'm in need of this variant for the page header work I've been doing so if you don't mind I can review tomorrow and merge when it's good to go?

davidlygagnon:
@kevinparkerson Go for it - sounds good to me.

components/button-group/index.jsx
```diff
@@ -102,6 +99,15 @@ const ButtonGroup = (props) => {
 			</fieldset>
 		);
 	}
+
+	if (props.variant === 'list') {
+		return (
+			<ul className={classNames('slds-button-group-list', props.className)}>
+				{children}
```
kevinparkerson: Since every child within ButtonGroup of variant `list` is meant to be wrapped in an `li` tag, please make this code do that here. Something like this:
```
{React.Children.map(props.children, (child) => (
	<li>{child}</li>
))}
```

kevinparkerson:
@pulkonet please create an example of the list variant in use. Put it in it's own file called `list-variant.jsx` beneath `__examples__`. Add it to both `__docs__/site-stories.js` and `__docs__/storybook-stories.jsx`. See [the list variant on SLDS](https://www.lightningdesignsystem.com/components/button-groups/#List) for what this example should look like. Be sure to use actual LDSR Button components for the children rather than the manual buttons seen in that example.

components/button-group/index.jsx
pulkonet: Done.

Issue #1850, opened 2019-03-16
Builder Header: Add component


garygong:
Implement the Builder Header component according to spec: https://www.lightningdesignsystem.com/components/builder-header/

garygong:
Here's what I'm thinking for the interface.  What do you think?

### BuilderHeader:

| Name               | Type      | Description                                                                             |
|--------------------|-----------|-----------------------------------------------------------------------------------------|
| assistiveText.icon | string    | Assistive text for the title icon                                                       |
| className          | string    | Additional class names                                                                  |
| children           | Element[] | Structure of the header. BuilderHeaderNav (optional) or BuilderHeaderToolbar (optional) |
| events.onClickBack | function  | Callback for when "Back" is clicked                                                     |
| events.onClickHelp | function  | Callback for when "Help" is clicked                                                     |
| pageType           | string    | Title of the builder                                                                    |

### BuilderHeaderNav:

| Name     | Type      | Description                 |
|----------|-----------|-----------------------------|
| children | Element[] | Dropdown or Button elements |

### BuilderHeaderToolbar:

| Name                  | Type          | Description                                           |
|-----------------------|---------------|-------------------------------------------------------|
| assistiveText.actions | string        | Assistive text for the actions section of the toolbar |
| children              | ButtonGroup[] | Button groups in the toolbar                          |
| onRenderActions       | functions     | Renders the actions section of the toolbar            |

### Example:
Here's an example using the example given in the spec.
```
<BuilderHeader
	assistiveText={{ icon: 'Builder' }}
	pageType="Page Type"
	title="App Name"
>
	<BuilderHeaderNav>
		<Button
			iconCategory="utility"
			iconName="settings"
			label="Link"
		/>
		<Dropdown
			iconCategory="utility"
			iconCategory="page"
			label="Dropdown"
		/>
	</BuilderHeaderNav>
	<BuilderHeaderToolbar>
		<ButtonGroup>
			<Button
				assistiveText={{ icon: 'Undo' }}
				iconCategory="utility"
				iconName="undo"
			/>
			<Button
				assistiveText={{ icon: 'Redo' }}
				iconCategory="utility"
				iconName="redo"
			/>
		</ButtonGroup>
		<ButtonGroup>
			<Button
				assistiveText={{ icon: 'Cut' }}
				iconCategory="utility"
				iconName="cut"
			/>
			<Button
				assistiveText={{ icon: 'Copy' }}
				iconCategory="utility"
				iconName="copy"
[...]

PR #1953, opened 2019-04-25
Builder Header: Add component

garygong:
I made some minor modifications to the proposed component API.  The changes are described here: https://github.com/salesforce/design-system-react/issues/1850#issuecomment-486856614

Screen Shot 2019-04-25 at 3 34 00 PM

@interactivellama When you get a chance, can you take a look?  Thanks!

Fixes https://github.com/salesforce/design-system-react/issues/1850


components/builder-header/__examples__/base-with-toolbar.jsx
```diff
@@ -0,0 +1,108 @@
...
+				helpIcon: 'Help',
+				icon: 'Builder',
+			}}
+			labels={{
+				back: 'Back',
+				help: 'Help',
+				pageType: 'Page Type',
+				title: 'App Name',
+			}}
+			style={{ position: 'relative' }}
```
garygong: This style override was needed because the header has `position: fixed`, so it was showing up at the top of the doc site.

components/builder-header/nav.jsx
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * The navigation section of the header.
+ */
+const BuilderHeaderNav = (props) => (
+	<nav className="slds-builder-header__item slds-builder-header__nav">
+		<ul className="slds-builder-header__nav-list">
+			{React.Children.map(props.children, (child) => {
+				if (child.type.displayName === BUILDER_HEADER_NAV_LINK) {
+					return child;
+				} else if (child.type.displayName.indexOf(MENU_DROPDOWN) !== -1) {
```
garygong: Turns of display name was something like `ClickOutside(Dropdown)` because of the HOC.  I think it's ok to use `indexOf`?

components/builder-header/index.jsx
```diff
@@ -0,0 +1,200 @@
...
+						<h1 className="slds-builder-header__item-label">
+							<span className="slds-truncate" title={labels.pageType}>
+								{labels.pageType}
+							</span>
+						</h1>
+					</div>
+					<div className="slds-builder-header__item slds-builder-header__utilities">
+						<div className="slds-builder-header__utilities-item">
+							<a
+								href="javascript:void(0);"
```
garygong: I wasn't sure if we wanted to pass in `<a>` attributes like href, target, etc. to use with links instead of the `onClick`.  I think it's ok to just have `onClick`?

components/builder-header/index.jsx
```diff
@@ -0,0 +1,200 @@
...
+	 * _Tested with snapshot testing._
+	 */
+	labels: PropTypes.shape({
+		back: PropTypes.string,
+		help: PropTypes.string,
+		pageType: PropTypes.string,
+		title: PropTypes.string,
+	}),
+	/**
+	 * Custom styles to be passed to the component
```
interactivellama: This should mention the node `.slds-builder-header_container`

components/builder-header/index.jsx
interactivellama: onClick is always best I think, since you can do a location change, but not the reverse.

components/builder-header/nav.jsx
interactivellama: I would recommend a sub component wrapper / HOC like `BuilderNavDropdown`--that way we can change Dropdown as needed, but still use the same prop interface. We do that with `GlobalNavigationDropdown` today.

garygong: Good idea!

garygong:
Good idea about the status indicators in the toolbar.  I'll go ahead and add them.

garygong:
@interactivellama Can you take a look at the changes made in this commit: https://github.com/salesforce/design-system-react/pull/1953/commits/ed1485bcafb86531c86ca3dc7b45566f009a8653  Thanks!

interactivellama:
Looks good! Thanks for being patient with me.

PR #1979, opened 2019-05-10
[UNSUPPORTED PROTOTYPE] Add direction HOC and context to support RTL in components

### LINKED ISSUE #1970: [UNSUPPORTED PROTOTYPE] RTL Support

**High-level impl proposal**

**Scope**
- Icon (use RTL sprite)
- ComboBox (flip alignment)

davidlygagnon:
Fixes #1970 

### Additional description

-  Adding `DirectionSettings` HOC that introduces new direction prop/context on components.
-  Fixing inline styles specific for RTL in `Combobox` component.

---


components/utilities/direction/direction-settings.jsx
```diff
@@ -0,0 +1,25 @@
+/* Copyright (c) 2015-present, salesforce.com, inc. All rights reserved */
```
davidlygagnon: Added this file under `components/utilities/direction/direction-settings.jsx` instead of `components/utilities/private/direction/direction-settings.jsx` since I didn't feel the need for `private` folder. Let me know if you think otherwise.

davidlygagnon:
Examples in storybook. Note that this is still using the LTR SLDS css and only switched the markup to RTL `dir='rtl'`. I think it is sufficient for the examples. 

Screen Shot 2019-05-10 at 2 48 55 PM
Screen Shot 2019-05-10 at 2 31 00 PM

components/combobox/combobox.jsx
```diff
@@ -101,6 +102,10 @@ const propTypes = {
 		PropTypes.object,
 		PropTypes.string,
 	]),
+	/**
+	 * Establishes directional context for component. Defaults to left-to-right.
+	 */
+	direction: PropTypes.oneOf([DIRECTIONS.LTR, DIRECTIONS.RTL]),
```
interactivellama: I propose not making this public at this time.

components/utilities/direction/index.jsx
```diff
@@ -0,0 +1,28 @@
+/* Copyright (c) 2015-present, salesforce.com, inc. All rights reserved */
+/* Licensed under BSD 3-Clause - see LICENSE.txt or git.io/sfdc-license */
```
interactivellama: I think we might want the URL to be `components/utilities/UNSAFE_direction/index.jsx`. Otherwise you'd never see the unsafe part example for the examples.

davidlygagnon:
I added better comments in the storybook examples specifying the reason why the styles may look off. I would rather not add a static version of the RTL in the storybook because the css might get out of sync with what is on SLDS master as they move on to new versions. Once they ship a static version of the RTL that we can consume directly from NPM, we can always revisit and consume it.

PR #2077, opened 2019-06-19
[UNSUPPORTED PROTOTYPE] Dropdown: Add right to left language support (RTL)

### LINKED ISSUE #1970: [UNSUPPORTED PROTOTYPE] RTL Support

andyouyang22:
Fixes #1970 

### Additional description

---


components/menu-dropdown/menu-dropdown.jsx
```diff
@@ -182,6 +183,10 @@ const propTypes = {
 		PropTypes.object,
 		PropTypes.string,
 	]),
+	/**
+	 * Establishes directional context for component. Defaults to left-to-right.
+	 */
+	direction: PropTypes.oneOf([DIRECTIONS.LTR, DIRECTIONS.RTL]),
```
minevskiy: I don't think Menu needs to know about `direction`

components/combobox/combobox.jsx
```diff
@@ -467,11 +467,9 @@ class Combobox extends React.Component {
 			? 'relative'
 			: this.props.menuPosition; // eslint-disable-line react/prop-types
 
-		const alignment =
-			this.props.direction === DIRECTIONS.RTL ? 'bottom right' : 'bottom left';
```
minevskiy: ComboBox doesn't need `direction` prop anymore.

components/combobox/combobox.jsx
```diff
@@ -307,7 +307,6 @@ const defaultProps = {
 		removePill: ', Press delete or backspace to remove',
 		selectedListboxLabel: 'Selected Options:',
 	},
-	direction: DIRECTIONS.LTR,
```
interactivellama: I'm just glancing at this, but I think removal would be a breaking change to a publicly released library. Feel free to add comments and create a removal issue though.

minevskiy: Makes sense, let's keep it.

components/menu-dropdown/__examples__/custom-trigger.jsx
```diff
@@ -20,6 +20,10 @@ class Example extends React.Component {
 						{ label: 'Menu Item Three', value: 'C0' },
 						{ type: 'divider' },
 						{ label: 'Menu Item Four', value: 'D0' },
+						{ label: 'Menu Item Five', value: 'E0' },
+						{ label: 'Menu Item Six', value: 'F0' },
+						{ type: 'divider' },
+						{ label: 'Menu Item Seven', value: 'G0' },
```
andyouyang22: Added more items so that the dropdown scrollbar will be visible in storybook examples

components/combobox/combobox.jsx
andyouyang22: Hmm are we expected to accommodate both (1) using the `UNSAFE_Direction.Provider` context provider and (2) inputting the `direction` prop directly? It seems like if we are only supporting the first use case, we can safely remove the `direction` prop in combobox.jsx

interactivellama: I would ask @davidlygagnon for the reason for this prop’s existence. 

If you all ONLY want language direction will be done by “global” Context instead of a wrapping HOC, we can remove the prop type (that is make it not a publicly documented API), but we need to keep the underlying code working and comment in the source that it should be removed in the future at a breaking change.

minevskiy: We do still intend to rely on `LanguageDirection` HOC to provide the `direction` prop value.
The reason this prop exists in `Combobox` is because `Combobox` used to change its dialog alignment based on direction. This is being generalized in this PR - the dialog alignment is handled by the Dialog itself. Therefore, we don't need the prop anymore, but we can keep it around until the next breaking change.

components/menu-dropdown/index.jsx
```diff
@@ -4,6 +4,7 @@
 // ### onClickOutside
 // Listen for clicks that occur somewhere in the document, outside of the element itself
 import onClickOutside from 'react-onclickoutside';
+import LanguageDirection from '../utilities/UNSAFE_direction/private/language-direction';
 import MenuDropdown from './menu-dropdown';
 
-export default onClickOutside(MenuDropdown);
+export default LanguageDirection(onClickOutside(MenuDropdown));
```
minevskiy: Don't even have to wrap  `MenuDropdown` with `LanguageDirection`. 
`Dialog` is [wrapping itself](https://github.com/salesforce/design-system-react/blob/master/components/utilities/dialog/index.jsx#L525).

components/menu-dropdown/menu-dropdown.jsx
```diff
@@ -39,6 +39,7 @@ import {
 	MENU_DROPDOWN_TRIGGER,
 	LIST,
 } from '../../utilities/constants';
+import { DIRECTIONS } from '../utilities/UNSAFE_direction';
```
minevskiy: unnecessary import

components/combobox/combobox.jsx
davidlygagnon: Yup - not needed anymore in this case.

components/menu-dropdown/index.jsx
```diff
@@ -4,6 +4,7 @@
 // ### onClickOutside
 // Listen for clicks that occur somewhere in the document, outside of the element itself
 import onClickOutside from 'react-onclickoutside';
+import LanguageDirection from '../utilities/UNSAFE_direction/private/language-direction';
```
minevskiy: unnecessary import

utilities/dialog-helpers.js
```diff
@@ -32,7 +32,18 @@ const mapPropToPopperPlacement = (propString) => {
 			placement = 'left-end';
 			break;
 		default:
-			placement = propString;
+			placement = align;
+	}
+	if (direction === 'rtl') {
```
minevskiy: Could we use `DIRECTIONS` enum here?

interactivellama:
Snapshots need updating.
Please @ reply me when you'd like a review.

andyouyang22:
@interactivellama Do you think you could take a quick look again? Thanks!

interactivellama:
Pushed up master branch

components/combobox/combobox.jsx
```diff
@@ -102,6 +102,10 @@ const propTypes = {
 		PropTypes.object,
 		PropTypes.string,
 	]),
+	/**
+	 * TODO - Remove this unused prop in the next major release
+	 */
+	direction: DIRECTIONS.LTR,
```
interactivellama: Go ahead and move this from the propTypes, and we can make it undocumented, but still be in the code.

components/combobox/__examples__/base-inherit-menu-width-RTL.jsx
```diff
@@ -86,7 +86,6 @@ class Example extends React.Component {
 					</h2>
 					<UNSAFE_DirectionSettings.Provider value="rtl">
 						<Combobox
-							isOpen
```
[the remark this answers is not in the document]
andyouyang22: Ah no, I thought this was left open by mistake

components/utilities/dialog/index.jsx
```diff
@@ -279,14 +281,11 @@ class Dialog extends React.Component {
 		// A Dropdown with overflowBoundaryElement position and 'align=right' uses max-width instead of inherited children width
 		const right = 'inherit';
 
-		const leftValue = this.props.direction === DIRECTIONS.RTL ? right : left;
-		const rightValue = this.props.direction === DIRECTIONS.RTL ? 0 : right;
```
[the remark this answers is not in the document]
andyouyang22: This logic was originally only working for Combobox and not MenuDropdown. The new logic in `mapPropToPopperPlacement` covers both cases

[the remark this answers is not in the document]
minevskiy: I also works for `Popover`.

minevskiy:
@interactivellama , could we please merge this?

PR #2089, opened 2019-06-20
[UNSUPPORTED PROTOTYPE] Datepicker: Add right to left language support (RTL)

### LINKED ISSUE #1970: [UNSUPPORTED PROTOTYPE] RTL Support

hajiz:
Fixes #1970 

### Additional description

Support RTL in date picker component. The work done is mostly around keyboard interaction that was flipped. Added some tests to the existing snapshot test for the component, since story book testing for date picker is disabled (refer to [story-based-tests.js](https://github.com/salesforce/design-system-react/blob/97443696da1c49409b370d7031a3b9c5f1560415/components/story-based-tests.js#L68))

---


components/date-picker/__docs__/storybook-stories.jsx
```diff
@@ -12,13 +12,26 @@ import CustomInput from '../__examples__/custom-input';
 import SnaphotDefault from '../__examples__/snapshot-default';
 import WeekdayPicker from '../__examples__/weekday-picker';
 
+// eslint-disable-next-line camelcase
+import UNSAFE_DirectionSettings from '~/components/utilities/UNSAFE_direction';
+
+const makeRtl = (component) => (
```
hajiz: I don't like how we are repeating makeRtl method. I will have another PR coming up for color picker. If no one objects I will move this to the direction/index.js and reuse.

@andyouyang22 @minevskiy

hajiz:
please review @minevskiy @interactivellama

components/date-picker/date-picker.jsx
```diff
@@ -59,6 +60,10 @@ const propTypes = {
 		PropTypes.object,
 		PropTypes.string,
 	]),
+	/**
+	 * Establishes directional context for component. Defaults to left-to-right.
+	 */
+	direction: PropTypes.oneOf([DIRECTIONS.LTR, DIRECTIONS.RTL]),
```
interactivellama: FYI: I’m not sure what direction we need to go, but @andyouyang22 and @minevskiy are talking about context only usage over on Dropdown component.

minevskiy: @hajiz , DatePicker itself doesn't need to know about direction. Its children do. The whole point of context is to avoid passing props through the hierarchy of components.
Please undo changes in this file and in `date-picker/index.jsx`, `calendar-wrapper.jsx`, `calendar.jsx`, and instead wrap the export of `date-picker/private/day.jsx` with `LanguageDirection`.

minevskiy: @interactivellama , I didn't mean to suggest to only use the context. I meant that, if a component doesn't need to change its look or behavior based on direction, the it needs neither `direction` prop nor `LanguageDirection` HOC.

hajiz: Gotcha, I was wondering about this too. I'll remove the passing of params and switch to wrapping the day component in language direction.

interactivellama: Thank you for clarifying. I stepped into the middle without understanding.

hajiz: Addressed in 4553397

Thanks Stephan for catching it and Ivan for pointing out the good practice.

components/date-picker/__docs__/storybook-stories.jsx
minevskiy: I'm fine with extracting it into a test helper, but not sure it should be in `src`

hajiz:
@interactivellama @kevinparkerson @davidlygagnon @garygong this PR is ready for final approval and merge. Can you please take a look?

components/date-picker/__docs__/storybook-stories.jsx
interactivellama: Yeah, I'd recommend something in `/test` folder. There are other helper files in there. `rtl-helpers.js.jsx` is good.

components/date-picker/__tests__/date-picker.browser-test.jsx
```diff
@@ -12,6 +12,16 @@ import Datepicker from '../../date-picker';
 import Input from '../../input';
 import KEYS from '../../../utilities/key-code';
 
+// eslint-disable-next-line camelcase
+import UNSAFE_DirectionSettings from '../../utilities/UNSAFE_direction';
+
+const makeRtl = (component) => (
```
interactivellama: Another reason to put it in `tests` folder

interactivellama:
Updating my comment. I'm going to go ahead and merge tonight.

minevskiy:
I was testing this and noticed that, since DatePicker creates its own Portal, there is no way for me as a consumer to wrap that portal into a Provider.
I think we have to wrap `DatePicker` into `LanguageDirection` and pass direction as a prop to `Calendar` and `Day`.
@interactivellama , what do you think?

interactivellama:
Just to end this thread in case anyone is watching. We ending up wrapping the portal in order to carry over the Language Direction context.

PR #2099, opened 2019-06-21
[UNSUPPORTED PROTOTYPE] Colorpicker: Add right to left language support (RTL)

### LINKED ISSUE #1970: [UNSUPPORTED PROTOTYPE] RTL Support

hajiz:
Fixes #1970

### Additional description

Changes are two folds, one keyboard navigation is flipped, and also the popover is anchored to the bottom right of the picker in RTL context.

---


components/utilities/UNSAFE_direction/index.jsx
```diff
@@ -25,5 +25,12 @@ DIRECTIONS.RTL = 'rtl';
...
+const makeRTL = (component) => (
+	// eslint-disable-next-line
+	<UNSAFE_DirectionSettings.Provider value="rtl">
+		<div dir="rtl">{component}</div>
+	</UNSAFE_DirectionSettings.Provider>
+);
+
 // eslint-disable-next-line camelcase
-export { UNSAFE_DirectionSettings as default, DIRECTIONS };
+export { UNSAFE_DirectionSettings as default, DIRECTIONS, makeRTL };
```
hajiz: Once #2089 and #2077 are merged I can rebase and update all duplicated `makeRtl` calls to refer to this centralized one. If you disagree and think we should just repeat it wherever needed please let me know.

components/color-picker/index.jsx
```diff
@@ -57,6 +59,10 @@ const propTypes = {
 	 * Unique ID for component.
 	 */
 	id: PropTypes.string,
+	/**
+	 * Establishes directional context for component. Defaults to left-to-right.
+	 */
+	direction: PropTypes.oneOf([DIRECTIONS.LTR, DIRECTIONS.RTL]),
```
hajiz: color picker component does need the `direction` prop to decide the anchor point of the pop over, but does it need to be an official prop? I can remove this from here so that it doesn't show up in the docs. Is that preferred?

CC @minevskiy @andyouyang22 @interactivellama

minevskiy: `Popover` alignment will be correctly flipped by its inner `Dialog` (see https://github.com/salesforce/design-system-react/pull/2077/). 
No need to adjust it here.

components/utilities/UNSAFE_direction/index.jsx
minevskiy: I don't think this should be in `src`. It's more of a test utility.

components/color-picker/index.jsx
hajiz: @minevskiy did you mean to put this comment on line 400?

I should wait until #2077 and #2099 are merged and then rebase and merge this one. Need that also for that makeRtl reuse

hajiz: But again regarding this line. If I add it here it gets added to `components/component-docs.json`. I'm wondering if we want that.

components/utilities/UNSAFE_direction/index.jsx
hajiz: Sure. I'll find a better place for it. It's also used in storybook code, other than tests.

components/color-picker/index.jsx
minevskiy: > did you mean to put this comment on line 400?
Since we don't need to flip Popover alignment, we also don't need `direction` prop.

hajiz: oh, lol, good point!! 👍

interactivellama:
@hajiz Please request a review when you would like one. I’ve removed image snapshots from the library. It was causing issues now that most of the components have them.

I’d recommend pushing up master.

hajiz:
Hi @interactivellama , since #2077 isn't merged yet I will pull master into this branch and undo my makeRtl change (plus the image diff). @andyouyang22 will then take care of unifying the makeRtl calls. I'll ping you once it's ready for a review. Thanks

components/color-picker/index.jsx
```diff
@@ -19,7 +19,6 @@ import TabsPanel from '../tabs/panel';
 import Popover from '../popover';
 
 import ColorUtils from '../../utilities/color';
-
```
hajiz: I meant no change to this file. If you want I can put this new line back

hajiz:
Thanks @minevskiy , and sorry I had forgot to remove the image snapshot I had added earlier.

hajiz:
Finally, all the tests are passing and the change is ready for final review. Thanks again Ivan and Stephan

minevskiy:
I'm afraid we have the same problem as with `DatePicker`. Since `ColorPicker` is creating its own portal, a consumer cannot wrap that portal into a provider. We might have to wrap `ColorPicker` into `LanguageDirection` instead and pass direction as a prop to `SwatchPicker`

minevskiy:
Maybe we could wrap portal contents into `UNSAFE_DirectionSettings.Provider` similar to how they are wrapped into `IconSettings`? And the provider value would come from the `direction` prop.
https://github.com/salesforce/design-system-react/blob/master/components/utilities/dialog/index.jsx#L508-L512
```
<Portal onOpen={this.handleOpen} portalMount={this.props.portalMount}>
	<UNSAFE_DirectionSettings.Provider value={this.props.direction}>
		<IconSettings {...truthyIconSettingsContext}>
			{contents}
		</IconSettings>
	</UNSAFE_DirectionSettings.Provider>
</Portal>
```

interactivellama:
Jumping in here quickly, I’m cool with the Provider in the portal. The library has tech debt in that it’s still not using the built in portals, I believe the built in portals from React would pass on the correct context without the HOC.

minevskiy:
That's right, the built-in Portals preserve context, but custom DSR Portals don't.
@hajiz , I think you can make the change either in this PR or in a new one.

hajiz:
@minevskiy @interactivellama the PR should be ready now! 🤞

interactivellama:
Looks good. Thanks @hajiz !

hajiz:
Thank you @interactivellama for merging the change

PR #2150, opened 2019-07-05
Add Expression Component

### LINKED ISSUE #1695: Expression: Props proposal
Expression builders help users declaratively construct logical expressions. These expressions can be used when querying for a filtered set of records, creating rules to control when something executes, or any other conditional logic.

https://www.lightningdesignsystem.com/components/expression/

aswinshenoy:
Fixes #1695 

---


kevinparkerson:
@aswinshenoy I think I like the declarative setup seen here, though it'd help to see a more interactive example. Could you rig up the base example so that "Add Condition" and "Add Group" actually works to see how complicated it could get?

components/expression/condition.jsx
```diff
@@ -0,0 +1,206 @@
...
+							<span>{this.props.labels && this.props.labels.condition ? this.props.labels.condition : null }</span>
+							<span className="slds-assistive-text">{this.props.assistiveText ? this.props.assistiveText.title : 'Condition' }</span>
+						</legend>
+						<div className="slds-grid slds-gutters_xx-small">
+							<div className="slds-col">
+								<Combobox
+									events={{
+										onSelect: (event, data) => {
+											// eslint-disable-next-line no-unused-expressions
+											typeof this.props.onChangeResource === "function" ? this.props.onChangeResource(event,data) : null;
```
kevinparkerson: Please change this to:
```
if (this.props.onChangeResource) {
    this.props.onChangeResource(event, data);
}
```
PropTypes already protects the expected type, so we just want to make sure it's there before running the function.

components/expression/condition.jsx
```diff
@@ -0,0 +1,206 @@
...
+									<span className="slds-form-element__label">&nbsp;</span>
+									<div className="slds-form-element__control">
+										<Button
+											variant="outline-brand"
+											iconCategory="utility"
+											iconName="delete"
+											iconVariant="border-filled"
+											onClick={() => {
+												// eslint-disable-next-line no-unused-expressions
+												typeof this.props.onDelete === "function" ? this.props.onDelete() : null;
```
kevinparkerson: Same as before, convert to:
```
if (this.props.onDelete) {
    this.props.onDelete();
}
```

components/expression/group.jsx
```diff
@@ -0,0 +1,223 @@
...
+				trigger = "any"
+		} else if(selection === '3'){
+			  trigger = "custom"
+		} else if(selection === '4'){
+			  trigger = "always"
+		} else if(selection === '5'){
+			trigger = "formula"
+		}
+		// eslint-disable-next-line no-unused-expressions
+		typeof this.props.onChangeTrigger === "function" ? this.props.onChangeTrigger(trigger) : null;
```
kevinparkerson: Same as before, convert to:
```
if (this.props.onChangeTrigger) {
    this.props.onChangeTrigger(trigger);
}
```

components/expression/group.jsx
```diff
@@ -0,0 +1,223 @@
...
+
+		const buttons = (<div className="slds-expression__buttons">
+			<Button
+				iconCategory="utility"
+				iconName="add"
+				iconPosition="left"
+				label={this.props.labels ? this.props.labels.addCondition : "Add Condition" }
+				onClick={() => {
+					// eslint-disable-next-line no-unused-expressions
+					typeof this.props.onAddCondition === "function" ? this.props.onAddCondition() : null;
```
kevinparkerson: Same as before, convert to:
```
if (this.props.onAddCondition) {
    this.props.onAddCondition();
}
```

components/expression/group.jsx
```diff
@@ -0,0 +1,223 @@
...
+			/>
+			{ !this.props.child ? (
+				<Button
+					iconCategory="utility"
+					iconName="add"
+					iconPosition="left"
+					label={this.props.labels ? this.props.labels.addGroup : "Add Group" }
+					onClick={() => {
+						// eslint-disable-next-line no-unused-expressions
+						typeof this.props.onAddGroup === "function" ? this.props.onAddGroup() : null;
```
kevinparkerson: Same as before, convert to:
```
if (this.props.onAddGroup) {
    this.props.onAddGroup();
}
```

components/expression/condition.jsx
```diff
@@ -0,0 +1,206 @@
...
+			operator = "equal"
+		} else if(selection === '2'){
+			operator = "notEqual"
+		} else if(selection === '3'){
+			operator = "greaterThan"
+		} else if(selection === '4'){
+			operator = "lessThan"
+		}
+		// eslint-disable-next-line no-unused-expressions
+		typeof this.props.onChangeOperator === "function" ? this.props.onChangeOperator(operator) : null;
```
kevinparkerson: Please remove the `eslint-disable-next-line` comment, then convert this to:
```
if (this.props.onChangeOperator) {
    this.props.onChangeOperator(operator);
}
```
PropTypes already protects the expected function type, so we just want to make sure it's there before running the function.

kevinparkerson:
@aswinshenoy don't forget to add the `docs.json` file with subcomponents as dependencies and add these to the `components/index.jsx` file.

components/expression/private/condition.jsx
```diff
@@ -0,0 +1,239 @@
...
+
+	onDelete: PropTypes.func,
+
+	value: PropTypes.string,
+	onChangeValue: PropTypes.func,
+
+	isChild: PropTypes.bool,
+};
+
+const operatorSelections = [
```
kevinparkerson: data can't be in a LDSR component or private component like this as the developer can't modify this code for their apps

interactivellama:
@kevinparkerson Are you thinking along the lines of having a tree/collection of data in the example container state and having logic in the children that parses the tree into JSX?

Something like this pseudo-code:
```
state.data = [
{ type: "condition", resource: '1', operator: "equals", value: "" }
{ type: "group", children: [
    {type: "condition" resource: '1', operator: "equals", value: "" }
    ]
];

<Expression>
    {state.data.map(object) => {
    if object.type = "condition"
    return
    (<ExpressionCondition
		operator={object.operator}
		{...etc}
	/>)
```

components/expression/__examples__/base.jsx
```diff
@@ -0,0 +1,273 @@
...
+
+	addCondition() {
+		const { conditions } = this.state;
+		const newCondition = {
+			isGroup: false,
+			resource: '',
+			operator: '',
+			value: '',
+		};
+		conditions.push(newCondition);
```
interactivellama: spread `{...}` and `array.concat()` are your friends here. You shouldn't need `update` then. See the tree for some examples. `push` mutates variables which is why `update` is needed here.

interactivellama:
@aswinshenoy This is coming along nicely. Looking forward to meeting to discuss this.

components/expression/formula.jsx
```diff
@@ -0,0 +1,143 @@
...
+											variant="icon"
+											iconCategory="utility"
+											iconName="help"
+											onClick={this.props.onClickHelp}
+										/>
+									</div>
+								</div>
+								<div className="slds-rich-text-editor__textarea slds-grid">
+									<div
+										aria-label="Compose formula"
```
kevinparkerson: This will need to use `props.labels.textEditor` instead. Default that value to `Compose formula`

components/expression/formula.jsx
```diff
@@ -0,0 +1,143 @@
...
+	operatorInput: PropTypes.node,
+
+	onClickHelp: PropTypes.func,
+	onClickCheckSyntax: PropTypes.func,
+};
+
+const defaultProps = {
+	labels: {
+		label: 'Formula',
+		checkSyntax: 'Check Syntax',
```
kevinparkerson: add a `textEditor` entry defaulted to 'Compose formula'

components/expression/formula.jsx
```diff
@@ -0,0 +1,143 @@
...
+	 */
+	className: PropTypes.oneOfType([
+		PropTypes.array,
+		PropTypes.object,
+		PropTypes.string,
+	]),
+
+	labels: PropTypes.shape({
+		label: PropTypes.string,
+		checkSyntax: PropTypes.string,
```
kevinparkerson: Add a `textEditor` label prop

components/expression/formula.jsx
```diff
@@ -0,0 +1,143 @@
...
+											}}
+											variant="icon"
+											iconCategory="utility"
+											iconName="help"
+											onClick={this.props.onClickHelp}
+										/>
+									</div>
+								</div>
+								<div className="slds-rich-text-editor__textarea slds-grid">
+									<div
```
kevinparkerson: Use the https://github.com/lovasoa/react-contenteditable instead of a straight div here, which works around contenteditable React wonkiness. Make sure you run `npm install react-contenteditable --save-dev` so it's added to package.json. Here's an idea of what it'd look like:

```
import React from 'react'
import ContentEditable from 'react-contenteditable'

// ...

class ExpressionFormula extends React.Component {
	constructor() {
		super()
		this.textEditorRef = React.createRef();
		this.state = {
			textEditorValue: 'Compose formula...'	// default is set here to preserve functionality if not controlled
		};
	};
	
	// ...

	handleTextEditorChange = (event) => {
		const textEditorValue = event.target.value;
		
		if (this.props.textEditorValue === undefined) {
			this.setState({ textEditorValue });
		}
		
		if (this.props.onTextEditorChange) {
			this.props.onTextEditorChange(event, { textEditorValue });
		}
	};

	render = () => {
		// ...
	
		return (
			<!-- etc -->
			<ContentEditable
				aria-label={this.props.labels.textArea}	// This should work but let me know if it doesn't. Also, don't leave this comment in the PR :D
				className="slds-rich-text-area__content slds-text-color_weak slds-grow"
				innerRef={this.textEditorRef}
				html={this.props.textEditorValue !== undefined ? this.props.textEditorValue : this.state.textEditorValue}
				onChange={this.handleTextEditorChange}
			/>
			<!-- etc -->
		);
	};
};

// ...
```

components/expression/formula.jsx
```diff
@@ -0,0 +1,143 @@
...
+		label: PropTypes.string,
+		checkSyntax: PropTypes.string,
+	}),
+
+	resourceCombobox: PropTypes.node,
+	functionCombobox: PropTypes.node,
+	operatorInput: PropTypes.node,
+
+	onClickHelp: PropTypes.func,
+	onClickCheckSyntax: PropTypes.func,
```
kevinparkerson: Add two props:
- `textAreaValue`, of `PropTypes.node`. Notice the `node` type here; that's because a rich text editor might have special formatting beyond what a string can provide

components/expression/__examples__/base.jsx
kevinparkerson: Another option is to do this:
`const conditions = JSON.parse(JSON.stringify(this.state.conditions));`, which guarantees conditions is a new array. Note that this only works if there's only primitives in your data.

kevinparkerson:
@aswinshenoy I know you were initially wanting to solidify the API but now that it's further along go ahead and add comments to all the public component prop definitions. Please point out any new props you've added that I might have missed. The `formula` component props you've proposed here are all good, just need the two additions I mentioned.

components/expression/formula.jsx
aswinshenoy: thank you for suggesting `react-contenteditable`, I was looking for directions here. :)

components/expression/__examples__/base.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import Example from './example';
```
kevinparkerson: @aswinshenoy since this example will be shown on the site, unfortunately we can't import `Example` like in the other storybook variants. Please copy the contents of `__examples__/example.jsx` into this file just for this variant. Apologies for not calling this out sooner, wasn't thinking about the site imports!

kevinparkerson:
@aswinshenoy I went ahead and updated the `base.jsx` example, rennaming it `resource-selected.jsx` and copying `example.jsx` into it so that it'll work on the site since this was my initial oversight 😄

components/expression/__examples__/example.jsx
```diff
@@ -0,0 +1,498 @@
...
+		this.setState({ conditions });
+	}
+
+	render() {
+		return (
+			<IconSettings iconPath="/assets/icons">
+				<Expression
+					id="expression-example"
+					events={{
+						onChangeTrigger: (val) => {
```
kevinparkerson: Since you've changed the data being passed, the parameters here need to be `(event, data)` with the log function afterwards passing both. Also, the state setter should say `this.setState({ triggerType: data.triggerType });`. See the `resource-selected.jsx` example if needed.

components/expression/__examples__/example.jsx
```diff
@@ -0,0 +1,498 @@
...
+								event,
+								eventName: 'Custom Logic Changed',
+								data,
+							});
+							this.setState({ customLogic: data.value });
+						},
+						onAddCondition: () => {
+							log({
+								action: this.props.action,
+								event,
```
kevinparkerson: Since there's no event variable here, set this to either an empty object or `null`

components/expression/__examples__/example.jsx
```diff
@@ -0,0 +1,498 @@
...
+								eventName: 'New Condition Added',
+								data: null,
+							});
+							this.addCondition();
+						},
+						onAddGroup: () => {
+							log({
+								action: this.props.action,
+								event,
```
kevinparkerson: Same here, either have the event attribute be an empty object or `null`

components/expression/__examples__/example.jsx
```diff
@@ -0,0 +1,498 @@
...
+									log({
+										action: this.props.action,
+										event,
+										eventName: `Formula updated in Text Editor`,
+										data,
+									}),
+								onClickCheckSyntax: () =>
+									log({
+										action: this.props.action,
+										event,
```
kevinparkerson: Same here, since there's no event variable here, set this to either an empty object or `null`

components/expression/__examples__/example.jsx
```diff
@@ -0,0 +1,498 @@
...
+										eventName: `Check Syntax Button Clicked`,
+										data: null,
+									}),
+								onClickHelp: () =>
+									log({
+										action: this.props.action,
+										event,
```
kevinparkerson: Same here, since there's no event variable here, set this to either an empty object or `null`

components/expression/__examples__/example.jsx
```diff
@@ -0,0 +1,498 @@
...
+													event,
+													eventName: `Condition ${i} Value Changed`,
+													data,
+												});
+												this.updateData(i, data.value, 'value');
+											},
+											onDelete: () => {
+												log({
+													action: this.props.action,
+													event,
```
kevinparkerson: Same here, since there's no event variable here, set this to either an empty object or `null`

components/expression/condition.jsx
```diff
@@ -0,0 +1,238 @@
...
+	 */
+	value: PropTypes.string,
+};
+
+const defaultProps = {
+	assistiveText: {
+		title: 'Condition',
+		deleteIcon: 'Delete Condition',
+	},
+	isSubCondition: false,
```
kevinparkerson: This default is not needed, since if it's not passed in it will be `undefined` and equate to falsey values in conditionals

components/expression/group.jsx
```diff
@@ -0,0 +1,279 @@
...
+	triggerType: PropTypes.oneOf(['all', 'any', 'custom', 'always', 'formula']),
+	/**
+	 * Sets the input for the custom logic value input box, shown if the `triggerType` is set to `custom`.
+	 */
+	customLogicValue: PropTypes.string,
+};
+
+const defaultProps = {
+	triggerType: 'all',
+	isRoot: false,
```
kevinparkerson: Same here, this default for `isRoot` isn't needed since it will evaluate as undefined and be falsey unless provided

PR #2181, opened 2019-07-17
Add Tree Grid Component

### LINKED ISSUE #1836: Add Tree Grid component
Add Tree Grid (https://www.lightningdesignsystem.com/components/tree-grid/).

aswinshenoy:
Fixes #1836 
---


components/tree-grid/__examples__/default.jsx
```diff
@@ -0,0 +1,133 @@
+import React from 'react';
+import TreeGrid from '~/components/tree-grid';
+import IconSettings from '~/components/icon-settings';
+
+const data = {
+	cols: [
```
interactivellama: The columns should be a separate prop. See DataTable component.

components/tree-grid/__examples__/default.jsx
```diff
@@ -0,0 +1,133 @@
+import React from 'react';
+import TreeGrid from '~/components/tree-grid';
+import IconSettings from '~/components/icon-settings';
+
+const data = {
```
interactivellama: If there were no nested rows, then a dev would use a DataTable. I don't think we need this example and base should be nested.

components/tree-grid/__examples__/nested.jsx
```diff
@@ -0,0 +1,156 @@
+import React from 'react';
+import TreeGrid from '~/components/tree-grid';
+import IconSettings from '~/components/icon-settings';
+
+const data = {
```
interactivellama: The columns should be child components. See `DataTable` component.

interactivellama: Here is the Lightning component data for instance.
```
({ // eslint-disable-line
    init: function (cmp) {
        var columns = [
            {
                type: 'text',
                fieldName: 'accountName',
                label: 'Account Name',
                initialWidth: 300
            },
            {
                type: 'number',
                fieldName: 'employees',
                label: 'Employees'
            },
            {
                type: 'phone',
                fieldName: 'phone',
                label: 'Phone Number'
            },
            {
                type: 'url',
                fieldName: 'accountOwner',
                label: 'Account Owner',
                typeAttributes: {
                    label: { fieldName: 'accountOwnerName' }
                }
            },
            {
                type: 'text',
                fieldName: 'billingCity',
                label: 'Billing City'
            }
        ];

        cmp.set('v.gridColumns', columns);

        // data
        var nestedData = [
            {
                "name": "123555",
                "accountName": "Rewis Inc",
                "employees": 3100,
                "phone": "837-555-1212",
                "accountOwner": "http://example.com/jane-doe",
                "accountOwnerName": "Jane Doe",
                "billingCity": "Phoeniz, AZ"
            },

            {
                "name": "123556",
                "accountName": "Acme Corporation",
                "employees": 10000,
                "phone": "837-555-1212",
                "accountOwner": "http://example.com/john-doe",
                "accountOwnerName": "John Doe",
                "billingCity": "San Francisco, CA",
                "_children": [
                    {
                        "name": "123556-A",
                        "accountName": "Acme Corporation (Bay Area)",
                        "employees": 3000,
                        "phone": "837-555-1212",
                        "accountOwner": "http://example.com/john-doe",
                        "accountOwnerName": "John Doe",
                        "billingCity": "New York, NY",
                        "_children": [
                            {
                                "name": "123556-A-A",
                                "accountName": "Acme Corporation (Oakland)",
                                "employees": 745,
                                "phone": "837-555-1212",
                                "accountOwner": "http://example.com/john-doe",
                                "accountOwnerName": "John Doe",
                                "billingCity": "New York, NY"
                            },
                            {
                                "name": "123556-A-B",
                                "accountName": "Acme Corporation (San Francisco)",
                                "employees": 578,
                                "phone": "837-555-1212",
                                "accountOwner": "http://example.com/jane-doe",
                                "accountOwnerName": "Jane Doe",
                                "billingCity": "Los Angeles, CA"
                            }
                        ]
                    },

                    {
                        "name": "123556-B",
                        "accountName": "Acme Corporation (East)",
                        "employees": 430,
                        "phone": "837-555-1212",
                        "accountOwner": "http://example.com/john-doe",
                        "accountOwnerName": "John Doe",
                        "billingCity": "San Francisco, CA",
                        "_children": [
                            {
                                "name": "123556-B-A",
                                "accountName": "Acme Corporation (NY)",
                                "employees": 1210,
                                "phone": "837-555-1212",
                                "accountOwner": "http://example.com/jane-doe",
                                "accountOwnerName": "Jane Doe",
                                "billingCity": "New York, NY"
                            },
                            {
                                "name": "123556-B-B",
                                "accountName": "Acme Corporation (VA)",
                                "employees": 410,
                                "phone": "837-555-1212",
                                "accountOwner": "http://example.com/john-doe",
                                "accountOwnerName": "John Doe",
                                "billingCity": "New York, NY",
                                "_children": [
                                    {
                                        "name": "123556-B-B-A",
                                        "accountName": "Allied Technologies",
                                        "employees": 390,
                                        "phone": "837-555-1212",
                                        "accountOwner": "http://example.com/jane-doe",
                                        "accountOwnerName": "Jane Doe",
                                        "billingCity": "Los Angeles, CA",
                                        "_children": [
                                            {
                                                "name": "123556-B-B-A-A",
                                                "accountName": "Allied Technologies (UV)",
                                                "employees": 270,
                                                "phone": "837-555-1212",
                                                "accountOwner": "http://example.com/john-doe",
                                                "accountOwnerName": "John Doe",
                                                "billingCity": "San Francisco, CA"
                                            }
                                        ]
                                    }
                                ]
                            }
                        ]
                    }
                ]
            },

            {
                "name": "123557",
                "accountName": "Rhode Enterprises",
                "employees": 6000,
                "phone": "837-555-1212",
                "accountOwner": "http://example.com/john-doe",
                "accountOwnerName": "John Doe",
                "billingCity": "New York, NY",
                "_children": [
                    {
                        "name": "123557-A",
                        "accountName": "Rhode Enterprises (UCA)",
                        "employees": 2540,
                        "phone": "837-555-1212",
                        "accountOwner": "http://example.com/john-doe",
                        "accountOwnerName": "John Doe",
                        "billingCity": "New York, NY"
                    }
                ]
            },

            {
                "name": "123558",
                "accountName": "Tech Labs",
                "employees": 1856,
                "phone": "837-555-1212",
                "accountOwner": "http://example.com/john-doe",
                "accountOwnerName": "John Doe",
                "billingCity": "New York, NY",
                "_children": [
                    {
                        "name": "123558-A",
                        "accountName": "Opportunity Resources Inc",
                        "employees": 1934,
                        "phone": "837-555-1212",
                        "accountOwner": "http://example.com/john-doe",
                        "accountOwnerName": "John Doe",
                        "billingCity": "Los Angeles, CA"
                    }
                ]
            }
        ];

        cmp.set('v.gridData', nestedData);


        var expandedRows = ["123556", "123556-A", "123556-B", "123556-B-B", "123557"];

        cmp.set('v.gridExpandedRows', expandedRows);
    }
}); // eslint-disable-line
```

interactivellama: The data should be very similar. I would change `_children` to `nodes` align with `Tree` component. Please review Tree data https://github.com/salesforce/design-system-react/blob/master/utilities/sample-data/tree/nodes-base.jsx#L2

components/tree-grid/__examples__/nested.jsx
```diff
@@ -0,0 +1,156 @@
...
+					href: 'javascript:void(0);',
+				},
+				{
+					label: 'Phoenix, AZ',
+				},
+			],
+		},
+		{
+			id: '2',
+			childOpen: true,
```
interactivellama: The `Tree` keys are:

```
	 *   expanded: string,
	 *   id: string,
	 *   label: string or node,
	 *   selected: string,
	 *   type: string,
	 *   nodes: array
```

components/tree-grid/__examples__/nested.jsx
```diff
@@ -0,0 +1,156 @@
...
+				},
+				{
+					label: '10000',
+				},
+				{
+					label: '837-555-1212',
+				},
+				{
+					label: 'Jane Doe',
+					href: 'javascript:void(0);',
```
interactivellama: The default cell render should be just a string. If you'd like to scope this project to only output that I'm cool with that.

`DataTable` has a `CustomCell` render that receives the tree data as props. The `CustomCell` is the child of the Column child. See https://github.com/salesforce/design-system-react/blob/master/components/data-table/__examples__/advanced.jsx#L10

components/tree-grid/index.jsx
```diff
@@ -0,0 +1,198 @@
+/* Copyright (c) 2015-present, salesforce.com, inc. All rights reserved */
+/* Licensed under BSD 3-Clause - see LICENSE.txt or git.io/sfdc-license */
+
+// Implements the [Welcome Mat design pattern](https://lightningdesignsystem.com/components/welcome-mat/) in React.
```
interactivellama: Typo

components/tree-grid/index.jsx
```diff
@@ -0,0 +1,198 @@
...
+		PropTypes.array,
+		PropTypes.object,
+		PropTypes.string,
+	]),
+	/**
+	 * HTML id for component.
+	 */
+	id: PropTypes.string,
+
+	labels: PropTypes.shape({
```
interactivellama: These aren't visible, so they go in assistiveText prop. Please see:
https://github.com/salesforce/design-system-react/blob/master/components/data-table/index.jsx#L55

components/tree-grid/index.jsx
```diff
@@ -0,0 +1,198 @@
...
+		return (
+			<table
+				id={this.getId()}
+				aria-multiselectable="true"
+				className={classNames(
+					'slds-table',
+					'slds-table_bordered',
+					'slds-table_edit',
+					'slds-table_fixed-layout',
+					'slds-table_resizable-cols',
```
interactivellama: Unless you are implementing resizable, this might need to be removed. I'd recommend not implementing it. It's not in the current Data either. Column width on the other hand is needed.

components/tree-grid/private/row.jsx
```diff
@@ -0,0 +1,133 @@
...
+									</a>
+								) : (
+									<>{col.label}</>
+								)}
+							</div>
+						</td>
+					)
+			)}
+			<td role="gridcell" style={{ width: '3.25rem' }}>
+				<Button
```
interactivellama: This component should accept a `TreeRowActions` component as a child.

interactivellama:
Typically, this component is going to have a fixed header, but if you want to descope that for now I'm OK with that. This comments this is OK: http://design-system-react-components.herokuapp.com/?path=/story/sldsdatatable--advanced-fixed-layout

but you don't have to do this: http://design-system-react-components.herokuapp.com/?path=/story/sldsdatatable--fixed-header

interactivellama:
I will try to get you a new props proposal by tomorrow, but feel free to ask questions tonight and maybe we can discuss tomorrow in the meeting.

interactivellama:
So far this is what I have:

```
/* 
Optional:
* Inline editing and sorting of columns are not supported by the Lightning Component https://developer.salesforce.com/docs/component-library/bundle/lightning:treeGrid/documentation
* Column options is not needed
*/

// ### Prop Types
const propTypes = {
  /**
   * **Assistive text for accessibility.**
   * This object is merged with the default props object on every render.
   * * `actionsHeader`: Text for heading of actions column
   * * `columnSort`: Text for sort action on table column header
   * * `columnSortedAscending`: Text announced once a column is sorted in ascending order
   * * `columnSortedDescending`: Text announced once a column is sorted in descending order
   * * `selectAllRows`: Text for select all checkbox within the table header
   * * `selectRow`: Text for select row
   */
  assistiveText: PropTypes.shape({
    actionsHeader: PropTypes.string,
    // columnSort: PropTypes.string,
    // columnSortedAscending: PropTypes.string,
    // columnSortedDescending: PropTypes.string,
    selectAllRows: PropTypes.string,
    selectRow: PropTypes.string
  }),
  /**
   * Provide children of the type `<DataTableColumn />` to define the structure of the data being represented and children of the type `<DataTableRowActions />` to define a menu which will be rendered for each item in the grid. Use a _higher-order component_ to customize a data table cell that will override the default cell rendering. `CustomDataTableCell` must have the same `displayName` as `DataTableCell` or it will be ignored. If you want complete control of the HTML, including the wrapping `td`, you don't have to use `DataTableCell`.
   * ```
   * import DataTableCell from 'design-system-react/data-table/cell';
   * const CustomDataTableCell = ({ children, ...props }) => (
   *   <DataTableCell {...props} >
   *   <a href="javascript:void(0);">{children}</a>
   *   </DataTableCell>
   * );
   * CustomDataTableCell.displayName = DataTableCell.displayName;
   *
   * <TreeGrid>
   *   <TreeGridColumn />
   *   <TreeGridColumn>
   *   <TreeGridCustomCell />
   *   </TreeGridColumn>
   *   <TreeGridRowActions />
   * </TreeGrid>
   * ```
   */
  children: PropTypes.node,
  /**
   * Class names to be added to the table.
   */
  className: PropTypes.oneOfType([
    PropTypes.array,
    PropTypes.object,
    PropTypes.string
  ]),
  // this function is how child rows can be fetched asyncronously from the server
  // For this project it can be the default and use `forceUpdate` if you would like. Otherwise, please review Tree and https://github.com/paularmstrong/normalizr This concept took me a while to understand, but the gist is that you flatten all the nested data into an array and link the data via object keys. This is how normalized databases work with tree data.
  getNodes: node => node.nodes,
  /**
   * A unique ID is needed in order to support keyboard navigation and ARIA support.
   */
  id: PropTypes.string,
  /**
   * The collection of items to render in the table. This is an array of objects with each object having keys that correspond with the  `property` prop of each `DataTableColumn`.
   */
  nodes: PropTypes.arrayOf(
    PropTypes.shape({
      id: PropTypes.string.isRequired
    })
  ).isRequired,
  /**
   * This function triggers when a row is expanded
   */
  onExpandRow: PropTypes.func.isRequired,
  /**
   * This function fires when the selection of rows changes. This component passes in `event, { selection }` to the function. `selection` is an array of objects from the `items` prop.
   */
  onRowChange: PropTypes.func,
  /**
   * Specifies a row selection UX pattern.
   * * `multiple`: This is the default
   * * `single`: Single row selection.
   * _This prop used to be a `boolean`, a `true` value will be considered `checkbox` for backwards compatibility._
   */
  selectRows: PropTypes.oneOf(["single", "multiple"]),
  /**
   * TreeGrids have horizontal borders by default. This removes them.
   */
  unborderedRow: PropTypes.bool
};

const test = (
  <TreeGrid
    assistiveText={{
      actionsHeader: "actions",
      columnSort: "sort this column",
      columnSortedAscending: "asc",
      columnSortedDescending: "desc",
      selectAllRows: "all rows",
      selectRow: "select this row"
    }}
    getNodes={this.getNodes()}
    id="TreeGridExample"
    onRowChange={this.handleChanged}
    onExpandRow={this.handleExpandRow}
    onSort={this.handleSort}
  >
    <TreeGridColumn
      label="Name"
      property="opportunityName"
    >
      <CustomTreeGridCell />
    </TreeGridColumn>
    <TreeGridColumn label="Account Name" property="accountName" />
    <TreeGridColumn label="Close Date" property="closeDate" />
    <TreeGridColumn label="Stage" property="stage" />
    <TreeGridColumn
      label="Confidence"
      property="confidence"
    />
    <TreeGridColumn label="Amount" property="amount" />
    <TreeGridColumn label="Contact" property="contact">
      <CustomTreeGridCell />
    </TreeGridColumn>
    <TreeGridRowActions
      options={[
        {
          id: 0,
          label: "Add to Group",
          value: "1"
        },
        {
          id: 1,
          label: "Publish",
          value: "2"
        },
        {
          id: 2,
          label: "Third of Seven",
          value: "3"
        },
        {
          id: 3,
          label: "Fourth of Seven",
          value: "4"
        },
        {
          id: 4,
          label: "Fifth of Seven",
          value: "5"
        },
        {
          id: 5,
          label: "Sixth of Seven",
          value: "6"
        },
        {
          id: 6,
          label: "Seventh of Seven",
          value: "7"
        }
      ]}
      menuPosition="overflowBoundaryElement"
      onAction={this.handleRowAction}
    />
  </TreeGrid>
);
```

interactivellama:
@aswinshenoy Please remove `TreeGrid` Column Actions (dropdown) from the scope of this component.

interactivellama:
Selecting a parent should not select the children. Checkbox and selection is for "row selection" only.

interactivellama:
As we discussed on Wed, "select all" checkbox should switch to Indeterminate state when only a few rows are selected.

interactivellama:
Please add a single select headless example.

components/tree-grid/docs.json
```diff
@@ -0,0 +1,7 @@
+{
+	"component": "tree-grid",
+	"status": "prod",
```
interactivellama: This should be `prototype` until I can get a full accessibility review by a specialist.

components/tree-grid/index.jsx
interactivellama: This is still outstanding. ^

interactivellama: oh, sorry, it's duplicate now. Please remove `chooseRow`, `selectAll`.

components/tree-grid/index.jsx
```diff
@@ -0,0 +1,293 @@
...
+
+	labels: PropTypes.shape({
+		chooseRow: PropTypes.string,
+		selectAll: PropTypes.string,
+	}),
+	/**
+	 * Specifies a row selection UX pattern.
+	 * * `multiple`: This is the default
+	 * * `single`: Single row selection.
+	 * _This prop used to be a `boolean`, a `true` value will be considered `checkbox` for backwards compatibility._
```
interactivellama: This legacy issue should only pertain to DataTable and not TreeGrid.

interactivellama:
If you look at https://design-system-react-co-pr-2181.herokuapp.com/?path=/story/sldstreegrid--default there is flash of icon reload when you expand a node. I forget what in the code causes this (I'll look into it), but the reason it happens is that the whole component is re-rendering. The only node re-rendering should be the node expanding.

When I originally wrote the Tree, it did the same thing. It doesn't now ( https://react.lightningdesignsystem.com/components/trees/ ) It has something to do with all the nodes changing/updating, instead of only the one that was expanded.

components/tree-grid/index.jsx
```diff
@@ -0,0 +1,293 @@
...
+									<div
+										className="slds-th__action slds-th__action_form"
+										tabIndex="-1"
+									>
+										<Checkbox
+											assistiveText={{
+												label: assistiveText.selectAll,
+											}}
+											name="options"
+											// indeterminate={
```
interactivellama: I think I might try an reducer function passed down that manipulates a `isSelectAll` state in `index.js` https://www.w3schools.com/jsref/jsref_reduce.asp

pseudo code: `selectAll = true if all rows rendered are selected` so on each row render you'd run the function and update the parent's checkbox state.

The main issue is then what do we do about non-rendered rows. Although I think we we pass an `onSelectAll` function out, it's up to the dev to update their data, so I'm ok, with select all checkbox only reflecting real data state. Does that makes sense?

In DataTable we did this, but it was an array.
```
		const indeterminateSelected =
			canSelectRows && numRows !== numSelected && numSelected !== 0;
```

@kevinparkerson What do you think?

interactivellama: I'm hesitant to `this.flattenTree` but maybe we should re-use that function since it already walks the tree.

interactivellama:
@aswinshenoy What do you need from me to complete this component?

interactivellama:
At a minimum, we need a test suite.

aswinshenoy:
@interactivellama I am working on implementing the indeterminate state for checkboxes. I am slightly confused with it.
As you suggested selecting a row shouldn't automatically select its `hidden` child nodes. Therefore, on selecting a row with children, it will be checked. However, if you expand it, none of its children are selected, so its a checked state + indeterminate state. If allowed propogation of selection to its children, this wouldnt be a issue, but since we are not, we should switch selection to indeterminate when expanded, which seems to me a bit awkward. What do you think?

aswinshenoy:
@interactivellama I am also slightly confused what to do `onSelectAll`, whether select all the nodes, or only the top level nodes? 
image

Here too, there is 2 cases - 
1. children are hidden when the selection happen, in this case, the above result is obtained on expanding the selected row. This looks somewhat okay.
2. children are visible when the selection happen, but even in this case, only the top level nodes are selected, and the expanded sub-rows are left as such (which doesn't look good to me).
A workaround solution can be selected all the **visible** rows, but I am not sure whether this would be a SLDS pattern. 
Perhaps, it should be noted that all such manipulations are happening only in the examples, and the component is a controlled component / stateless. Therefore, I guess, the consumers are free to change the logic as they wish, and we shouldn't be worrying to much?

aswinshenoy:
In the single select variant, I have used `onClick` of the row for selection, but that causes the expansion button inside the row (`onExpand`) to also trigger causing expansion & selection together. 

It would be better if could adopt the Datatable single select design pattern for Treegrid against the current SLDS pattern.
image

aswinshenoy:
I am still unable to fix the issue of `moreActionsDropdown` passed in of column getting overlapped by the rows. I tried playing around with z-index, but doesn't seem to work :( Can you please guide me on this @interactivellama?

image

aswinshenoy:
I have used event trap for selection and expansion handlers, and everything seems to be working fine now. In case of the checkbox, I had made the example exactly as you had said, so it didn't require any changes. 

I am yet to touch the switching part between Navigation Mode <-> Selection Mode. As far as I learnt: -
1. Pressing Right once should expand
2. Pressing Right once again should switch mode to navigation
3. Pressing Left in the left-most cell on navigation mode will switch it back to selection mode

However, I have a slight confusion, regarding what objects will be focused
a. should all columns be focused/hovered one after other using is-hovered ? - https://www.lightningdesignsystem.com/components/tree-grid/#Item-Hovered
b. should only the columns with actions like links allowed to be hovered? (I don't know how we will identify it though)

I think we would be going with hovering each and every column in row (maybe except columns with some `hoverDisabled` prop)?

PR #2206, opened 2019-08-02
Added icon prop to Scoped Notification

### LINKED ISSUE #2203: ScopedNotification component needs an `icon` property
Currently, the ScopedNotification component accepts `iconName`, which doesn't allow much customization of the icon category, style, className, etc. Adding an `icon` property which accepts an Icon component would solve this. I will be doing this shortly as it is a relatively simple task.

kevinparkerson:
Fixes #2203

Added icon prop to scoped notification component and deprecated iconName prop

---


interactivellama:
DOM snapshots are off. Seems like it's missing assistive text maybe?

kevinparkerson:
@interactivellama they aren't failing locally... super weird

kevinparkerson:
@interactivellama ScopedNotification is used in SetupAssistant, but I forgot to update the DOM snapshots for that one. There is default assistive text now so that's what was different.

components/scoped-notification/index.jsx
```diff
@@ -21,24 +24,31 @@ const propTypes = {
...
 class ScopedNotification extends React.Component {
+	constructor(props) {
+		super(props);
+		checkProps(SCOPED_NOTIFICATION, props, componentDoc);
+	}
+
 	render() {
+		// For backwards compatibility
+		const iconAssistiveText =
+			(this.props.assistiveText && this.props.assistiveText.icon) || 'Info';
```
interactivellama: I'm not seeing `this.props.assistiveText.icon` in the propTypes and am confused.

I'm thinking if the prop is undefined, nothing should be present. In my original comment, I was saying that the snapshot wasn't passing. I wasn't saying that the previous snapshot was correct.

This seems like it should be a defaultValue prop and not defined in the render function. What's the reason for it not being a default prop?

kevinparkerson: @interactivellama I was struggling with what to do with this... in the original code, `this.props.assistiveText.icon` was incorrectly passed to the `Icon` prop as-is (`Icon` only accepts a `label` assistiveText option) but not listed in PropTypes or in defaultProps. Now that we have an `icon` prop, assistiveText should be passed in to the `Icon` component itself, which would make `this.props.assistiveText.icon` redundant... but we still have an automatically generated info Icon with assistiveText even if the `icon` prop is not set so I suppose that should still be exposed 😕

I'm thinking the following should work:
- Expose `assistiveText.icon` in the `PropTypes` so that people are aware it's there but don't set a defaultValue
- For cases where `props.icon` is present, clone the `Icon` component and it's props with an assistiveText override that merges the Icon assistiveText object on top of a compatible ScopedNotification assistiveText object
- For cases where `props.icon` is not present, assistiveText will be determined by `(this.props.assistiveText && this.props.assistiveText.icon) || 'Info'` as before

I'll push an update with that approach and we can decide whether to move forward with it or iterate again then.

kevinparkerson:
@interactivellama take a look at that solution for the assistiveText problem. I made sure to add a comment mentioning that the `Icon` component's `label` assistiveText would override the Scoped Notification assistiveText

interactivellama:
This approach is fine. In general, its unnecessary to merge two props into the sub components, but I can see how an engineer would like this approach based on the rest of the components.

PR #2219, opened 2019-08-05
Accordion: Add arrow key support

### LINKED ISSUE #2218: Accordion: Add arrow key navigation between headers
In the ARIA Authoring practices, it recommends the ability to use up/down arrow keys to cycle focus through the accordion panel headers (but does not automatically open the panel).

https://www.w3.org/TR/wai-aria-practices-1.1/#accordion

shleewhite:
Fixes #2218 

### Additional description

---


shleewhite:
@interactivellama the travis test failed because it timed out waiting for network availability. Can it be re-run?

interactivellama:
Restarted.

components/accordion/index.jsx
```diff
@@ -45,17 +45,55 @@ const propTypes = {
...
+	}
+
 	componentWillMount() {
 		this.generatedId = shortid.generate();
 	}
 
+	componentDidUpdate(prevState) {
+		if (this.state.currButtonIndex !== prevState.currButtonIndex) {
+			const { id } = this.props.children[this.state.currButtonIndex].props;
+			document.getElementById(`${id}-accordion-button`).focus();
```
interactivellama: Please do not use DOM query selection here. Please use React refs. An alternative approach would be by passing a function down into the Panel (`refs.summaryButton`) and the the panel passing a ref from the render function back up into the parent. I think the refs would be stored in an array. Here is radio's input ref for example: https://github.com/salesforce/design-system-react/blob/master/components/radio/index.jsx#L107

Two other more complicated options are:
* The other option would be to render each panel with a `hasFocus=true` prop and that typically involves a second flag that says whether any of the buttons are currently focused. The button would only focus if both flags were true, so that we only get one `focus` call. Datepicker does this type of approach: https://github.com/salesforce/design-system-react/blob/master/components/date-picker/private/calendar.jsx#L82
* The way that the Combobox menu items work is with an `active == props.activeOptionIndex` compare. You'd pass the active panel down.

components/accordion/panel.jsx
```diff
@@ -30,6 +30,10 @@ const propTypes = {
 	 * Component that can be passed as prop to `<Panel />`. As an example, a menu dropdown could be used here to handle additional actions for each accordion panel. _Tested with Mocha framework._
 	 */
 	panelContentActions: PropTypes.node,
+	/**
+	 * Callback that will run whenever there is a keydown on the panel button. Function doesn't change the state of the component.
+	 */
+	onKeyDown: PropTypes.func,
```
interactivellama: I'd recommend scoping this to something like `onKeyDownSummary`

Issue #2450, opened 2020-01-27
AccordionPanel crashes when no `refs` prop


burtek:
If `<AccordionPanel>` gets no `refs` prop, it throws 

```
Uncaught TypeError: Cannot read property 'summaryButton' of undefined
    at SLDSAccordionPanel
```

due to this line:
https://github.com/salesforce/design-system-react/blob/a4092d9013792edea5b64e079fa99c177972ff02/components/accordion/panel.jsx#L83

According to [docs](https://react.lightningdesignsystem.com/components/accordions/#sub-components), `refs` is not required though.

Workaround: add `refs={{}}` to props, but it's pretty irritating.

sumukhah:
working on the issue,
thank you

sumukhah:
this issue will not occur if you have used `<AcordianPanel />` as a child of `<Acordion />`
But it would be nice to fix this  So that one can use AcordianPanel without Acordion.

burtek:
@sumukhah I used `<AccordionPanel />` as child of `<Accordion />` and it crashed.

burtek:
ping (un-stale)

PR #2452, opened 2020-01-30
Fix AccordionPanel: application crashes if refs is not provided

sumukhah:
refs is not a required prop for the AccordinPanel component, but if we have not provided refs, refs will be undefined,
refs.summaryButton will not exist. This makes the app crash. This error will occur only if we use AccordinPanel
individually(Not as a child of Accordin). That is because Accordin component will pass refs by default.

This issue has fixed using a ternary expression. If refs exist then buttonRef takes refs.summaryButton otherwise
it will have a null value. Along with this, I have removed onClickSummary prop. as it is unused, throughout the component

Fixes: #2450


sumukhah:
Hello,
I need some help to figure out what went wrong in my code.
It would be really helpful if somebody points out what I made mistake..

burtek:
@sumukhah doesn't look like your fault.

@interactivellama will this get merged or sth?

rainerbruemmer:
@sumukhah How did you solve your problem (without a merged pull request)? I have the same problem.

sumukhah:
@rainerbruemmer, Manually copy and paste my pull request code in your code in respective files.

rainerbruemmer:
> @rainerbruemmer, Manually copy and paste my pull request code in your code in respective files.

Ok, thank you. But that is not satisfying.

@interactivellama @shleewhite Are there any plans to merge this PR?

interactivellama:
I'm cool with testing refs and returning `null`. `currButtonIndex` is used many, many times. Non-SLDS patterns are not robustly supported. What is the use case for an `AccordionPanel` outside an `Accordion`? Is there a link on the SLDS site you could share? https://www.lightningdesignsystem.com/

burtek:
@interactivellama in my case it was that `AccordionPanel`s were grouped in `React.Fragment` for some reason (can't remember why now and I don't work there any more). See https://github.com/salesforce/design-system-react/issues/2450#issuecomment-580265827

rainerbruemmer:
> What is the use case for an `AccordionPanel` outside an `Accordion`?
I implemented a react component which renders an `AccordionPanel` and contains some logic and an own state. So my `AccordionPanel` is use inside a `Accordion`, but wrapped in another component.

Monday I will replace my specialised `AccordionPanel` with the plain `AccordionPanel` and do the special things in a component which is given as a child. This solution is more clean and reusable (and hopefully working)

interactivellama:
That sounds like an HOC. Maybe there's a way to pass the instance up?

https://github.com/christianalfoni/formsy-react/issues/308

Issue #2490, opened 2020-03-28
Filter and Panel markup needs to be updated


garygong:
I noticed that the Filter close button was out of date with what is in the latest version of SLDS:
https://lightningdesignsystem.com/components/expression/#Filtering

We have an "x" icon when it should be a trash can icon.

I did a few spot checks for other places in the markup that might need to be updated, and noticed that there were some differences.  We should make sure that the Filter and Panel markup is up-to-date with SLDS as part of fixing this issue.

I will work on this.

PR #2504, opened 2020-04-10
Filter and Panel markup updates

garygong:
Some minor updates to Filter and Panel markup to bring them up-to-date with the latest from SLDS.

The only visible change is the "x" icon changing into a trash can icon:

Screen Shot 2020-04-10 at 3 10 08 PM

Fixes https://github.com/salesforce/design-system-react/issues/2490

@interactivellama Can you please review?  Thanks!

### DIFF
--- a/components/panel/filtering/list-heading.jsx
+++ b/components/panel/filtering/list-heading.jsx
@@ -61,7 +61,7 @@ PanelFilterListHeading.propTypes = {
 
 PanelFilterListHeading.defaultProps = {
 	heading: 'Matching all these filters',
-	lockedLabel: 'Locked filters',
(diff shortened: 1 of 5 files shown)


components/panel/filtering/list-heading.jsx
```diff
@@ -61,7 +61,7 @@ PanelFilterListHeading.propTypes = {
 
 PanelFilterListHeading.defaultProps = {
 	heading: 'Matching all these filters',
-	lockedLabel: 'Locked filters',
+	lockedHeading: 'Locked filters',
```
garygong, line 64: This was a typo that was causing the default value to be ignored.

PR #2506, opened 2020-04-14
Added support for aria-* to StatefulButton.

### LINKED ISSUE #2502: Accepting and supporting aria-* in StatefulButton similar to Button
I would like to add support for aria-* attributes in \<StatefulButton/\>.   Currently \<Button/\> already supports **aria-*** (https://github.com/salesforce/design-system-react/blob/master/components/button/index.jsx#L335).  But \<StatefulButton/\> has it hard coded to **aria-live="polite"** (https://github.com/salesforce/design-system-react/blob/master/components/button-stateful/index.jsx#L202).

This causes problem for our customers because it does not allow customization.  Their accessibility team has different requirements and want aria-pressed in place of aria-live so the state can be read by reader.

So I would like to add the support to StatefulButton, similar to Button, such that we can customize to be compliant with our customers' accessibility requirements.

The implementation can simply mirror what Button is currently doing, using the getAriaProps() from here https://github.com/salesforce/design-system-react/blob/master/utilities/get-aria-props.js to parse props, and then add valid **aria-*** as attributes in render().

To make sure the change is backward compatible, **aria-live="polite"** will be used as before, if no aria-* is provided as props.

kmmkwong:
This is to add support for aria-* in StatefulButton, similar to Button's support for aria-. To preserve the original behavior, the default aria-live attribute will be added if no aria- is provided in props.

Fixes #2502

### Additional description

---


kmmkwong:
> Thanks for the contribution! It looks like @kmmkwong is an internal user so signing the CLA is not required. However, [we need to confirm this](https://docs.google.com/forms/d/e/1FAIpQLSfB_vIrwFLweXIGt7ID98JhISCU2zm2ErOugAnGAcCvAzr7NQ/viewform).

I am not able to get pass this issue, even after submitting the form and re-creating the PR.  Can anyone help?

kmmkwong:
Anyone can help with code review here?

components/button-stateful/index.jsx
```diff
@@ -196,10 +199,18 @@ class ButtonStateful extends React.Component {
 
 		const isActive = typeof active === 'boolean' ? active : this.state.active;
 
+		// Accept aria-* props
+		let ariaProps = getAriaProps(this.props);
+
 		if (variant === 'icon' || variant === 'icon-filled') {
+			// Default aria attribute for stateful button with icon, if none is specified
+			if (!ariaProps || Object.keys(ariaProps).length === 0) {
```
garygong, line 207: I think `!ariaProps` here could be removed because `getAriaProps` will always return an object.  The occurrence down below too?

components/button-stateful/index.jsx
```diff
@@ -125,6 +126,8 @@ const defaultProps = {
 /**
  * The ButtonStateful component is a variant of the Lightning Design System Button component. It is used for buttons that have a state of unselected or selected.
  * For icon buttons, use <code>variant='icon'</code>. For buttons with labels or buttons with labels and icons, pass data to the state props (ie. <code>stateOne={{iconName: 'add', label: 'Join'}}</code>).
+ * Although not listed in the prop table, all `aria-*` will be added to the button element if passed in.
+ * If no `aria-*` is being passed, <code>aria-live='polite'</code> will be used for `icon` and `icon-filled` variant <code>aria-live='assertive'</code> will be used for other variants.
```
garygong, line 130: Can you update this comment to be a little clearer:

> Although not listed in the prop table, all `aria-*` props will be added to the button element if passed in.
> 
> If no `aria-*` props are passed in, <code>aria-live='polite'</code> is used for `icon` and `icon-filled` variants, and <code>aria-live='assertive'</code> is used for the remaining variants.
```

File: components/button-stateful/__docs__/storybook-stories.jsx [hunk 3]
```diff
@@ -24,4 +24,7 @@ storiesOf(BUTTON_STATEFUL, module)
 	.add('Base', () => getButtonStateful())
 	.add('Disabled', () => getButtonStateful({ disabled: true }))
 	.add('Icon Text Button', () => <IconTextButton />)
-	.add('Icon Button', () => <IconButton />);
+	.add('Icon Button', () => <IconButton />)
+	.add('Base With aria-pressed', () =>
```
garygong, line 28: This example is kind of confusing because `aria-pressed` is always true regardless of the state of the button.  Maybe use it as a controlled component so that this can be a more complete example?

components/button-stateful/__tests__/button-stateful.browser-test.jsx
```diff
@@ -11,10 +11,13 @@ chai.should();
 
 describe('Button Stateful: ', () => {
 	// Base defaults
-	const requiredProps = {
+	const requiredPropsNoVariant = {
```
garygong, line 14: To avoid creating an addition variable here, you could go with the original 
```
const requiredPropsNoVariant = {
	assistiveText: { icon: 'like' },
	iconName: 'like',
	iconSize: 'large',
	variant: 'icon'
};
```
and then switch the `assign` call arguments below so that the custom props overwrite values in `requiredProps`.

components/button-stateful/index.jsx
kmmkwong: I clarified a bit, but this is mostly the same as Button which already supports this.

components/button-stateful/__tests__/button-stateful.browser-test.jsx
kmmkwong: I think it is more readable and maintainable to keep the 2 set of required values, and keep them fixed.  In fact, the switch in the assign() was a deliberate test fix because the overwritten value ended up interfering each other in the test runs when I was extending the tests.

components/button-stateful/__docs__/storybook-stories.jsx
kmmkwong: Yeah you are right.  I removed this example, and instead, extended the IconButton example to include the toggling of the aria attribute according to the state of the button.

garygong:
LGTM!

PR #2663, opened 2020-10-22
Add infinite loading to data table

### LINKED ISSUE #2642: Support infinite scrolling with fixed column headers
DataTable should support infinite scrolling with fixed column headers.

My proposal is to follow similarly to how [lightning:datatable](https://developer.salesforce.com/docs/component-library/bundle/lightning:datatable/example) implemented infinite scrolling.

Provide these new props:
Property | Type | Description
-------- | ---- | -----------
hasMore | boolean | Specifies whether there's more data to be loaded and displays a spinner at the bottom of the table if so. The default is false.
onLoadMore | function | The action triggered when infinite loading loads more data.
loadMoreOffset | number | Determines when to trigger infinite loading based on how many pixels the table's scroll position is from the bottom of the table. The default is 20.

This solution will not do any window rendering but will allow the consumer to add additional rows to the items array.

Prophetess:
Fixes #2642

### Additional description

---

### DIFF
--- a/components/data-table/index.jsx
+++ b/components/data-table/index.jsx
@@ -386,6 +386,7 @@ class DataTable extends React.Component {
 			const action = [`${attach ? 'add' : 'remove'}EventListener`];
 			if (canUseEventListeners) {
 				window[action]('resize', this.resizeFixedHeaders);
+				window[action]('resize', this.onScrollerScroll);
 			}
 			if (canUseEventListeners && this.scrollerRef) {
 				this.scrollerRef[action]('scroll', this.resizeFixedHeaders);
--- a/components/data-table/index.jsx
+++ b/components/data-table/index.jsx
@@ -138,6 +138,10 @@ class DataTable extends React.Component {
 		 */
 		fixedLayout: PropTypes.bool,
 		/**
+		 * When fixedHeader is true, specifies whether there's more data to be loaded and displays a spinner at the bottom of the table if so.
+		 */
+		hasMore: PropTypes.bool,
+		/**
 		 * A unique ID is needed in order to support keyboard navigation and ARIA support.
 		 */
 		id: PropTypes.string,
@@ -157,6 +161,10 @@ class DataTable extends React.Component {
 		 */
 		joined: PropTypes.bool,
 		/**
+		 * Determines when to trigger infinite loading based on how many pixels the table's scroll position is from the bottom of the table.
+		 */
+		loadMoreOffset: PropTypes.number,
+		/**
 		 * A variant which removes hover style on rows
 		 */
 		noRowHover: PropTypes.bool,
@@ -164,6 +172,12 @@ class DataTable extends React.Component {
 		 * By default this function resizes the display headers when fixedHeader is `true`, but this behavior can be overridden. Passes an event and a data object with properties `headerRefs`, an array of DOM nodes referencing the `thead th` elements and `scrollerRef`, a DOM node referencing `.slds-table_header-fixed_scroller`
 		 */
 		onFixedHeaderResize: PropTypes.func,
+		/**
+		 * This function fires when infinite loading loads more data.
+		 *
+		 * This will be called multiple times while the table is being scrolled within the `loadMoreOffset`.  Please track whether or not loading is in progress and check it at the start of this function to avoid executing your callback too many times.
+		 */
+		onLoadMore: PropTypes.func,
 		/**
 		 * This function fires when the selection of rows changes. This component passes in `event, { selection }` to the function. `selection` is an array of objects from the `items` prop.
 		 *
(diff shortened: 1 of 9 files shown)


components/data-table/index.jsx
```diff
@@ -342,6 +359,19 @@ class DataTable extends React.Component {
...
 
+	onScrollerScroll = () => {
+		if (this.props.hasMore && this.props.onLoadMore) {
+			const scroller = this.scrollerRef;
+			if (
+				scroller &&
+				scroller.scrollTop + scroller.offsetHeight >
+					scroller.scrollHeight - this.props.loadMoreOffset
+			) {
+				this.props.onLoadMore();
```
garygong, line 370: It's possible that this callback is invoked multiple times as the user scrolls within the `loadMoreOffset` area.  Is that ok?  The caller may not expect to have this invoked repeatedly.

Would it make more sense to avoid repeatedly calling it until the next time the table height changes?

components/data-table/index.jsx
```diff
@@ -212,6 +217,18 @@ class DataTable extends React.Component {
 		 * A variant which removes horizontal padding. CSS class will be removed if `fixedLayout==true`.
 		 */
 		unbufferedCell: PropTypes.bool,
+		/**
```
garygong, line 220: Could you organize the props alphabetically?

components/data-table/index.jsx
```diff
@@ -451,82 +483,103 @@ class DataTable extends React.Component {
...
+									);
+							  })
+							: // Someday this should be an element to render when the table is empty
+							  null}
+					</tbody>
+				</table>
+				{this.props.fixedHeader && this.props.hasMore && (
+					<div
+						style={{
+							height: '3rem',
```
garygong, line 566: Instead of inline styles, it would be better if we could find SLDS classes that we could use, like in [Grid](http://lightningdesignsystem.com/utilities/grid/).

components/data-table/index.jsx
```diff
@@ -451,82 +483,103 @@ class DataTable extends React.Component {
...
+											rowActions={RowActions}
+											tableId={this.getId()}
+										/>
+									);
+							  })
+							: // Someday this should be an element to render when the table is empty
+							  null}
+					</tbody>
+				</table>
+				{this.props.fixedHeader && this.props.hasMore && (
```
garygong, line 563: Is there a need to require fixedHeader?  Would it be possible to use this without fixedHeader?

Prophetess: i don't think it would make sense since without fixed headers, it just renders the entire table without scrolling

components/data-table/index.jsx
Prophetess: Yes currently it would be called many times if the user keeps scrolling within the `loadMoreOffset`. I was considering an additional flag for the consumer to pass in to indicate it is currently loading but that flag would not affect any rendering of the component so thought it wasn't appropriate.  I thought we would just leave it to the consumer to handle multiple calls.

If the only call this callback the first time until the table height changes, what if the height doesn't change and they still want to load more (i.e. like in an error case that failed to load more).

garygong: What do you think about invoking it only when it crossing the `loadMoreOffset` threshold?  We'd have to keep track of the previous scroll position.

Prophetess: Sure I could do that... then it would only be called if they scrolled up then scrolled back down... is that any better?

components/data-table/index.jsx
Prophetess: There wasn't any SLDS markup for this so I couldn't see how they styled this.  I was trying to look at other components to see how they styled things but didn't see much usage of slds class names for things like this.

components/data-table/index.jsx
garygong: How about something like:
```
<div class="slds-grid slds-gutters">
  <div class="slds-col">
    <span>spinner</span>
  </div>
</div>```

File: components/data-table/index.jsx (hunk 3 above)
Comment (garygong) [reply components/data-table/index.jsx:566]: How about something like:
```
<div class="slds-grid slds-gutters">
  <div class="slds-col">
    <span>spinner</span>
  </div>
</div>```

components/data-table/index.jsx
garygong: I think it's _slightly_ better.  You're right that multiple calls are still possible, and that we should definitely mention that in the comments for the prop.

Prophetess: I've update the comment to be:

This function fires when infinite loading loads more data.

This will be called multiple times while the table is being scrolled within the `loadMoreOffset`.  Please track whether or not loading is in progress and check it at the start of this function to avoid executing your callback too many times.

garygong (approved):

garygong:
LGTM!

Issue #2723, opened 2021-01-05
Focus correct Combobox listbox option


garygong:
When clicking on a Combobox listbox option:

1. `focus` event is triggered (handlePillFocus): Focus changes to the active option (defaults to the first option)
2. `click` event is triggered (handlePillClickSelectedListbox): Focus changes to the clicked option

In an extreme case when some options are offscreen, this causes the wrong option to be focused and brought into view during (1), and the clicked option is never focused.

The fix is should be to set the correct active option during (1) so that focus always goes to the correct option.

PR #2724, opened 2021-01-05
Focus correct Combobox listbox option

garygong:
When clicking on a Combobox listbox option:
1 `focus` event triggered (handlePillFocus): Focus changes to the active option (defaults to the first option)
2 `click` event triggered (handlePillClickSelectedListbox): Focus changes to the clicked option

In an extreme case when some options are offscreen, this causes the wrong option to be focused and brought into view during (1), and the clicked option is never focused.

This fix sets the correct active option during (1) so that focus always goes to the correct option.

Fixes https://github.com/salesforce/design-system-react/issues/2723


alexxpan (approved):

alexxpan:
Is `handlePillClickSelectedListbox` still needed?

garygong:
You're right, since click also focuses, there's no need to have that callback at all.

alexxpan (approved):

PR #2790, opened 2021-03-09
Fix infinite scroll when items are shorter than table height

### LINKED ISSUE #2789: Infinite scrolling is broken when first page of results is shorter than the table height
If the height of all items in the first page of results is less than the height of the table container, the infinite scrolling breaks. This happens because the container is not scrollable, which prevents the `scroll` handler from being triggered.

jacobroschen:
Fixes #2789

### Additional description

This fixes the above issue by simulating a scroll event when the component is updated. This ensures that the data table fetches enough pages to allow the table table to scroll.

---

* [X] First-time contributors should sign the Contributor License Agreement. It's a fancy way of saying that you are giving away your contribution to this project. If you haven't before, wait a few minutes and a bot will comment on this pull request with instructions.

### DIFF
--- a/components/data-table/index.jsx
+++ b/components/data-table/index.jsx
@@ -153,7 +153,7 @@ class DataTable extends React.Component {
      *    </th>
      *  </tr>);
      *CustomHeaderRow.displayName = DataTableCell.displayName;
-     * 
+     *
      * <DataTable items=[{
         type: 'header-row',
         id: 'header-row-example-id-3',
@@ -198,7 +198,7 @@ class DataTable extends React.Component {
 		/**
 		 * This function fires when infinite loading loads more data.
 		 *
-		 * This will be called multiple times while the table is being scrolled within the `loadMoreOffset`.  Please track whether or not loading is in progress and check it at the start of this function to avoid executing your callback too many times.
+		 * This will be called multiple times while the table is being scrolled within the `loadMoreOffset`. It'll also continue to be called while `hasMore` is `true` and the table has not yet loaded enough rows to allow a user to scroll.  Please track whether or not loading is in progress and check it at the start of this function to avoid executing your callback too many times.
 		 */
 		onLoadMore: PropTypes.func,
 		/**
@@ -282,6 +282,10 @@ class DataTable extends React.Component {
 	componentDidUpdate() {
 		if (this.props.fixedHeader) {
 			this.resizeFixedHeaders();
+			// If the first page of results isn't enough to allow the user to scroll it causes
+			// the user to get into a state where they cannot load the second page.
+			// Simulating a scroll here will ensure that enough rows are loaded to enable scrolling
+			this.loadMoreIfNeeded();
 		}
 	}
 
@@ -388,7 +392,7 @@ class DataTable extends React.Component {
 		}
 	};
 
-	onScrollerScroll = () => {
+	loadMoreIfNeeded = () => {
 		if (this.props.hasMore && this.props.onLoadMore) {
 			const { scrollTop, offsetHeight, scrollHeight } = this.scrollerRef;
 			if (scrollTop + offsetHeight > scrollHeight - this.props.loadMoreOffset) {
@@ -411,11 +415,11 @@ class DataTable extends React.Component {
 			const action = [`${attach ? 'add' : 'remove'}EventListener`];
 			if (canUseEventListeners) {
 				window[action]('resize', this.resizeFixedHeaders);
-				window[action]('resize', this.onScrollerScroll);
+				window[action]('resize', this.loadMoreIfNeeded);
 			}
 			if (canUseEventListeners && this.scrollerRef) {
 				this.scrollerRef[action]('scroll', this.resizeFixedHeaders);
-				this.scrollerRef[action]('scroll', this.onScrollerScroll);
+				this.scrollerRef[action]('scroll', this.loadMoreIfNeeded);
 			}
 		}
 	};
--- a/components/component-docs.json
+++ b/components/component-docs.json
@@ -5238,7 +5238,7 @@
                 "returns": null
             },
             {
-                "name": "onScrollerScroll",
+                "name": "loadMoreIfNeeded",
                 "docblock": null,
                 "modifiers": [],
                 "params": [],
--- a/components/data-table/__tests__/data-table.browser-test.jsx
+++ b/components/data-table/__tests__/data-table.browser-test.jsx
@@ -11,6 +11,7 @@ import DataTableColumn from '../../data-table/column';
 import DataTableRowActions from '../../data-table/row-actions';
 import DataTableHighlightCell from '../../data-table/highlight-cell';
 import IconSettings from '../../icon-settings';
+import { mountComponent } from '../../../tests/enzyme-helpers';
 
 chai.should();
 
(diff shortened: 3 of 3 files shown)


jacobroschen:
@interactivellama it seems like you're still the most knowledgeable about DSR, do you have a few minutes to review this PR for me? Let me know if you need more info on the issue itself.

components/data-table/index.jsx
```diff
@@ -259,6 +259,10 @@ class DataTable extends React.Component {
 	componentDidUpdate() {
 		if (this.props.fixedHeader) {
 			this.resizeFixedHeaders();
+			// If the first page of results isn't enough to allow the user to scroll it causes
+			// the user to get into a state where they cannot load the second page.
+			// Simulating a scroll here will ensure that enough rows are loaded to enable scrolling
+			this.onScrollerScroll();
```
garygong, line 265: Could this have an unwanted effect of triggering an `onLoadMore()` call when the scroll position is outside the `loadMoreOffset` area?  For example, when there is an unrelated prop change like `selection` changes?

garygong: Is it possible for the consumer to calculate the right number of items and to provide them on initial render?  Or do we need to have DataTable repeatedly request items until there are enough?

jacobroschen: 1) It will call `onScrollerScroll()` when an unrelated prop changes, but it will only trigger `onLoadMore()` if the the table would need to load more given the current scroll position.
2) I think it would be extremely difficult for the consumer to compute the correct number of items to load on the initial render because it would need to reach into the inside of the DataTable to get the height, and then compute the height of each row. That feels like it would be exposing too much inside logic to the outside consumer, creating a more fragile DataTable.

Prophetess: You could calculate the height yourself outside of the DataTable using another container surrounding the DataTable to prefetch the appropriate number of items to fill the table.  That way you wouldn't need to know the inside logic.

jacobroschen: @Prophetess wouldn't you then need to duplicate the entire logic inside of onScrollScrolled in each consumer?

I don't see a case where we wouldn't want this to be the default behavior. Right now, infinite scrolling is broken on larger screens, and I don't see the need to make _every_ consumer to implement this internal logic to prevent this bug.

garygong: Instead of doing this check in `componentDidUpdate`, what do you think of moving this to `componentDidMount`, and call `onLoadMore` with the number of rows needed to fill the remaining space?  Something like:
```
componentDidMount() {
	if (hasSpaceForMoreRows) {
		const numRows = calculateRowsToFillAvailableSpace();
		this.props.onLoadMore(numRows);
	}
}
```
That way, the consumer knows how many rows are needed to fill the available space; plus, we wouldn't need to do the check on every prop update.

Prophetess: I would also prefer it being called only on mount instead of on every update.  Also it would be good to maybe have a prop `loadMoreOnMount` or something for consumers to be able turn on this behavior in case they won't want it called.

jacobroschen: @garygong I'm not sure moving the logic to `componentDidMount` would actually fix the bug we are having. Some consumers (including us) create the infinite scrolling data table with `items` set to an empty array. We do this because the data we put into the table is fetched asynchronously from an api, but we want to show the user the table with a spinner in the meantime (a perfectly normal UX pattern). Since the array is empty at the beginning, the data table would call the `onLoadMore` prop, however since we are already fetching the first page, another fetch wouldn't be triggered.

here's what the sequence of events would look like
- Start loading data table items from API
- Create infinite scrolling Data Table with `items={[]}`
- Data Table calls `onLoadMore`
- Our app sees we are already loading the first page, do nothing with the `onLoadMore` call since it would be fetching the same data (in fact, this is what the docs recommend <sup>1</sup>)
- Finish loading data table items from API
- Update Data Table with new items
- Nothing??? (unless this current change is implemented)
- User still cannot scroll because there aren't enough items

I also don't believe the `onLoadMore(numRows: number)` would be a good fix for two reasons. First, a lot of API's have a limit on how large of page sizes can be requested. Secondly, we'd still be forcing the consumer of this library to understand why `onLoadMore` has a `numRows` property, and why they need to load at least that many rows. It makes the API for this complex component even more complex.

If y'all are concerned with the the amount of work that is being done in `componentDidUpdate`, then you don't want to look at the `resizeFixedHeaders()` method. It calls `getBoundingClientRect()` in a loop, which in some browsers trigger a layout calculation.

1: `onLoadMore` docs - `This will be called multiple times while the table is being scrolled within the loadMoreOffset. Please track whether or not loading is in progress and check it at the start of this function to avoid executing your callback too many times.`

jacobroschen: That seems like it could work, but would it really be better than this solution where we can solve it at the library level for everyone automatically? I guess I really don't see the huge hesitation towards adding a small check on `componentDidUpdate` (which won't even be ran if they aren't using fixed headers/infinite scrolling).


> Existing consumers of the API can ignore numRows and would be unaffected

If by unaffected you mean that they'll still have this bug in their application, yeah.

jacobroschen: Now that I think about the `onLoadMore(numRows: number)`, we couldn't reliably calculate `numRows` for the consumer. We can't compute it because the height of each row is variable. Look at Custom Classes on the documentation page https://react.lightningdesignsystem.com/components/data-tables/

garygong: Yeah, the hesitation is that DataTable is a heavily used component with a history of performance issues.  So anything we can do to avoid unnecessary work should be done.  One possible solution for handling custom rows with unknown height is to provide the space in pixels, and allow the consumer to decide how to fill that space; and `numRows` would represent the standard row height.

If this seems too involved, we could go with the current check, but to only do it when infinite loading is enabled.  Ideally, there would be a prop for this like `enableInfiniteLoading`, but we can't introduce that now because that would break existing consumers.  Maybe we can use `onLoadMore` being specified as an indicator that infinite loading is enabled?

jacobroschen: `loadMoreIfNeeded` already does two checks to ensure that unnecessary work isn't done, `if (this.props.hasMore && this.props.onLoadMore) {`. From my understanding of the code and docs, `hasMore` and `onLoadMore` are the props needed to enable infinite scroll.

Prophetess: I am also concerned with breaking existing behavior.  When I implemented it I made it so that `onLoadMore` is triggered by a user action: scrolling. So I am not expecting it to call `onLoadMore` on an update... having this called now when it updates will trigger an unnecessary fetch for me.

Prophetess: Would it be possible to create a new callback prop instead of reusing the existing `onLoadMore` prop?

jacobroschen: @Prophetess It won't call `onLoadMore` unless more rows are actually needed to fill the space. So in your case, this change shouldn't impact you. I would appreciate if you could pull this change down to confirm that though.

Prophetess: Sure if it only calls `onLoadMore` when needed I'm fine with calling it in the update.  But the documentation for `onLoadMore` should be updated then to let consumers know of this behavior.

jacobroschen: @Prophetess I've updated the documentation. Please let me know if you can think of better wording.

garygong: Sorry for the delay, thanks a lot!

garygong (approved):

PR #2835, opened 2021-04-20
Combobox: Add select none pattern to read only variant (aka Picklist)

### LINKED ISSUE #2828: Add deselect/clear option to Picklist / Combobox Read-only pattern
I'm not recommending the following, but it gives you an idea of what I'm proposing for a read-only text.
Screen Shot 2021-04-15 at 5 03 20 PM
Screen Shot 2021-04-15 at 5 00 04 PM

Props proposal:
`hasFirstOptionDeselect`
`labels.firstOptionDeselect` defaults to `{blank}`

The UX pattern would be that the first option would never show with a checkbox and the value would revert to `undefined`. Therefore it would be a mixture of the above states.

### LINKED ISSUE #2843: Combobox: Down arrow key shows label of first menu item
Pressing down should highlight pre-selected item. The active menu index is correct, but the value/label is not updated.
See the following GIF:

interactivellama:
Fixes #2828
Fixes #2843

There is no SLDS pattern for this on the website. Selection of none, passes empty value and does not pass "none" to the `onChange` event.

### Additional description

Demo:

---


components/combobox/combobox.jsx
```diff
@@ -615,6 +623,16 @@ class Combobox extends React.Component {
 			options.push(...localProps.optionsSearchEntity);
 		}
 
+		if (props.hasDeselect) {
+			// eslint-disable-next-line fp/no-mutating-methods
```
swilliamsui, line 627: instead of continuing the anti pattern that requires adding the eslint exception why not create methods that create the options object with the additional option. that way the individual methods can be united tested.

components/combobox/combobox.jsx
```diff
@@ -615,6 +623,16 @@ class Combobox extends React.Component {
 			options.push(...localProps.optionsSearchEntity);
 		}
 
+		if (props.hasDeselect) {
```
swilliamset, line 626: Two observations:
1. instead of continuing the anti pattern requiring the eslint consider writing a method that returns an extended options list with the expected deselect object
2. this condition is also seen in the menu.jsx meaning there is opportunity to dry this up

If a new shared method is created then a single unit test against the shared method can be created.

components/combobox/combobox.jsx
```diff
@@ -1004,8 +1025,11 @@ class Combobox extends React.Component {
 			!this.props.multiple && !isSelected;
 		const multiSelectAndSelectedWasNotClicked =
 			this.props.multiple && !isSelected;
+		const deselectWasClicked = option.id === `${this.getId()}-deselect`;
```
swilliamset, line 1028: The string template 
```
`${this.getId()}-deselect`
```
is used more than once and could be dried with a method.

This would also help prevent errors if the suffix `-deselect` ever needs to change for some reason.

components/combobox/combobox.jsx
interactivellama: `options` array has been updated to not mutate

components/combobox/combobox.jsx
interactivellama: Class property added to make it DRYer.

garygong (approved):
LGTM

interactivellama:
Thanks Gary!

Issue #2860, opened 2021-05-05
DatePicker: Add support for error state


garygong:
Add support for error state to DatePicker: https://lightningdesignsystem.com/components/datepickers/#Date-Picker-Has-Error

PR #2861, opened 2021-05-05
DatePicker: Add support for error state

garygong:
Add a new prop `hasError` to `<DatePicker>` for the error state:

image

Fixes https://github.com/salesforce/design-system-react/issues/2860


components/date-picker/date-picker.jsx
```diff
@@ -85,6 +85,10 @@ const propTypes = {
 	 * Value of input that gets passed to `parser` prop on initial render. This prop is only present for uncontrolled use of Datepicker which is _highly discouraged_. A better name for this prop would be `defaultFormatedValue`. Please use the `value` prop instead. _Not tested._
 	 */
 	formattedValue: PropTypes.string,
+	/**
+	 * Applies the error style to the component.
+	 */
+	hasError: PropTypes.bool,
```
Azmarie, line 91: is `hasError` a required prop?

garygong: No, it's optional and will be interpreted as false when not passed in.

Azmarie (approved):

Azmarie (approved):

PR #3146, opened 2024-07-10
Update `combobox.jsx` based on WCAG Success Criteria: 4.1.2

### LINKED ISSUE #3144: @W-15466811 - Accessibility Issue: Year Combo Box Incorrectly Marked as Read-Only, Affecting Screen Reader Users
**Issue Description**

The combo box for selecting a year is incorrectly using the read-only attribute, which is intended for input fields to indicate they are not editable. This misuse violates the WCAG 2.1 success criteria 4.1.2 (Name, Role, Value), as it misleads screen reader users about the field's functionality, potentially causing confusion and hindering accessibility. Screen reader users are affected by this issue. 

Note: This issue has also been raised internally as a customer bug: [W-15466811](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001oDBl7YAG/view).

**Actual Result**

The year combo box is marked as read-only, implying it cannot be edited. However, the combo box is interactive, and users should be able to select a year from the dropdown list. Screen readers may incorrectly convey this as a non-interactive element, leading to user confusion.

**Expected Result**

Remove the read-only attribute from the year combo box to ensure it is correctly identified as an interactive element. This adjustment will align the combo box's behavior with its intended functionality, providing a more accessible experience for all users, including those relying on screen readers.

**WCAG Success Criteria**

4.1.2: Name, Role, Value

Impact

[Task Blocking]

```html
<div class="slds-form-element slds-align-content-center">
  <div class="slds-form-element">
    <label class="slds-form-element__label slds-assistive-text" for="fMQCKINvQE-year-picklist">Year</label>
    <div class="slds-form-element__control">
      <div class="slds-combobox_container">
        <div class="slds-combobox slds-dropdown-trigger slds-dropdown-trigger_click ignore-react-onclickoutside slds-shrink-none" role="combobox" aria-expanded="false" aria-haspopup="listbox">
          <div class="slds-combobox__form-element slds-input-has-icon slds-input-has-icon_right" role="none">
            <input autocomplete="off" class="slds-input slds-combobox__input" id="fMQCKINvQE-year-picklist" placeholder="Select an Option" readonly="" role="textbox" type="text" aria-autocomplete="list" value="2024">
            <span class="slds-icon_container slds-input__icon slds-input__icon_right">
              <svg aria-hidden="true" class="slds-icon slds-icon_x-small slds-icon-text-default" viewBox="0 0 52 52">
                <path d="M8.3 14h35.4c1 0 1.7 1.3.9 2.2L27.3 37.4c-.6.8-1.9.8-2.5 0L7.3 16.2c-.7-.9-.1-2.2 1-2.2z"></path>
              </svg>
            </span>
          </div>
        </div>
      </div>
    </div>
  </div>
</div>
```
June 5 Screenshot from Gus

**Findings**

The year selector combo box is configured as a read-only variant, which is causing accessibility tools to classify the picklist as read-only. This misconfiguration might be related to the specific styling needs of the year picklist. Refer to the [ComboBox documentation](https://github.com/salesforce/design-system-react/blob/master/components/date-picker/private/year-picklist.jsx#L93) for more details on the styling of each variant.

When the component is unpacked, this read-only attribute is added to the HTML div, causing accessibility tools to classify the picklist as read-only.

**Additional Notes**

I have also posted the same issue on the [#design-system-react-open-source](https://salesforce-internal.slack.com/archives/C122U6T4P/p1717425418562389?thread_ts=1717064600.908539&cid=C122U6T4P) Slack channel for further discussion.

anunay999:
Fixes #3144

### Additional description

---


components/combobox/combobox.jsx
```diff
@@ -1774,7 +1774,8 @@ class Combobox extends React.Component {
 								}
 							}}
 							placeholder={labels.placeholderReadOnly}
-							readOnly
+							// based on WCAG Success Criteria: 4.1.2: Name, Role, Value - input element is interactive
+							// readOnly
```
xulingzhihou, line 1778: please remove this extra comment

anunay999: Done

xulingzhihou:
@anunay999 I tested out your change in storybook. I think removing readOnly from input might not be the right fix for this issue and here is why:
1.  If a text input is not readOnly, it will show a cursor when focused indicating you can edit it, which is not what we want here.
image
2. I think the right solution would be to use a div or span instead of an input here. (See here: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/)
3. I noticed some other issues with how combobox is implemented. For example we have a container that wraps around the input widget that has `role="combobox"` instead of the input having role as combobox. 

I think combobox might requires a factoring to follow the most up to date accessibility standard.

interactivellama:
@anunay999 You may want to check out this blueprint. https://www.lightningdesignsystem.com/components/combobox/#Base-Combobox

I'm pretty sure the blueprints used to use `read-only`. It looks like the LWC version of this still uses `read-only`, too. https://developer.salesforce.com/docs/component-library/bundle/lightning-combobox/example

When reviewing the Base Combobox blueprint, keep in mind it may be using a different ARIA specs: 1.0 vs 1.1 vs 1.2.

The latest version of SLDS is using a span: `<span class="slds-truncate" id="combobox-value-id-96">Select an Option…</span>`

anunay999:
> @anunay999 You may want to check out this blueprint. https://www.lightningdesignsystem.com/components/combobox/#Base-Combobox
> 
> I'm pretty sure the blueprints used to use `read-only`. It looks like the LWC version of this still uses `read-only`, too. https://developer.salesforce.com/docs/component-library/bundle/lightning-combobox/example
> 
> When reviewing the Base Combobox blueprint, keep in mind it may be using a different ARIA specs: 1.0 vs 1.1 vs 1.2.
> 
> The latest version of SLDS is using a span: `<span class="slds-truncate" id="combobox-value-id-96">Select an Option…</span>`

Thank you for pointing out the differences in ARIA specifications and HTML structures between the SLDS and LWC versions of the combobox. Upon reviewing the LWC combobox example, it is evident that it uses a <span> element with the class slds-truncate to display the selected value. In contrast, the React implementation includes a readonly attribute on the <input> element.

---
Please find the attached screenshots from the above mentioned source for reference

**HTML structure for lwc**
Screenshot 2024-07-22 at 5 41 32 PM
---
**HTML structure for react**
Screenshot 2024-07-22 at 5 42 44 PM

interactivellama:
The `read-only` attribute was present to prevent folks from typing in the input. An input without the `read-only` is confusing for screen readers also, because the user will think they can type in it.

The LWC is using a button and the SLDS blueprint is using a `div` with the Combobox 1.2 spec (I believe). It's likely the only way to remove the read-only is to convert the Combobox to 1.2--we considered doing this about 5 years ago when the SLDS blueprint changed. After discussion the issue, the Analytics Cloud engineers agree to leave it the same, because change the spec would change the markup enough that it would be a breaking change.

In short, I'm totally open to updating to the 1.2 spec, but I don't think having a `role=textbox` with an `input` that is not `read-only` is a good idea either.


Also, here's the copy of the markup I'm seeing for the Base component which _is_ using `read-only`. It looks different than yours:
Screenshot 2024-08-13 at 1 20 48 AM

PR #3152, opened 2024-08-26
Fixed a runtime error caused by the Pills Container component trying to access an undefined icon.

### LINKED ISSUE #3138: The Pills Container component tries to access a property of undefined (runtime major crash)
**Error summary**: the [Pills Component](https://react.lightningdesignsystem.com/components/pill-container/) causes a major crash / runtime error

**Where the error occurs**:
design-system-react\components\pill-container\private\selected-listbox.jsx
at **line 149**, function **getIcon**

**Error description**: The function getIcon (pic 1) seems to handle the icon display for the currently selected option. Option === current selection, which legitimately can be undefined according to the component props (pic 2). The code tries to access the property option.icon and this results in an error "Cannot access property icon of undefined"

**Proposed solution**: since both selection and icon are optional props (pic 2), either change the props typings OR use optional chaining (?.) operator in the code. From the looks of it, the latter is preferable since the code handles the possible undefined everywhere except this one place. Seems like just a simple oversight but causes a big trouble in the app
image
image

shadwar123:
Fixes #3138

### Additional description

The Pills Container component was encountering a runtime error because it tried to access an icon that was sometimes undefined. To fix this, I applied optional chaining where the icon is being updated. If the icon is not present, we set its value to null. As a result, the getAvatar() function is called, and the props are passed with icon = null and avatar = value. In the Utilities, there's a Pill component that renders either the icon or the avatar based on their values using render() function.
@interactivellama 

---


interactivellama:
Thanks for picking this up! Please remove the package lock file changes unless you are updating dependencies.

components/pill/index.jsx
```diff
@@ -198,7 +198,7 @@ class Pill extends React.Component {
 	};
 
 	renderIcon = () => {
-		const icon = this.props.icon || this.props.avatar;
+		const icon = this.props?.icon || this.props?.avatar;
```
interactivellama, line 201: What is the use case for `this.props` not being defined?

shadwar123: Actually the undefined problem is comming beacuse of selected-listbox.jsx getIcon funtion as In some cases option only have icon or in other case only avatar, there i have applied const iconObject = option?.icon || null; and here const icon = this.props?.icon || this.props?.avatar; for the extra safety.

interactivellama:
I see that you downgraded the package lock verison. Please remove your package.lock file.

shadwar123:
> I see that you downgraded the package lock verison. Please remove your package.lock file.

Done,
Actually to run the program i have to run this npm install --legacy-peer-deps which changes Package-lock.json file. 
Now I have removed the Package-lock.json file.

interactivellama:
@shadwar123 You shouldn't delete package-lock files in libraries you are contributing to. Please restore the old version.

shadwar123:
> @shadwar123 You shouldn't delete package-lock files in libraries you are contributing to. Please restore the old version.

Done.

interactivellama (approved):

PR #3153, opened 2024-09-04
@W-16255012 : Updated combobox.jsx for combobox to open and close on enter key pressed

aarthim1111:
The combobox should be accessible via the enter key pressed just like how its accessible with the down key pressed.


aarthim1111:
@runpipeline@

aarthim1111:
@runpipeline@

PR #3195, opened 2026-09-10
feat: Modernize to React 19 + current toolchain (Vite 8, Vitest 4, ESLint 10, TypeScript 7)

Niznikr:
## Summary

Modernizes `@salesforce/design-system-react` to **React 19** with a current toolchain (Vite 8, Vitest 4, ESLint 10 flat config, TypeScript 7, Storybook 10, Node 24), migrates the test suite from Enzyme/Mocha to Vitest + React Testing Library, incrementally converts components to TypeScript, and removes deprecated dependencies. Along the way it fixes the React-19 removed-API breakages and a number of component/Storybook bugs surfaced by the upgrades.

> **Note on size:** this is a large, long-running modernization branch. It is organized into reviewable, self-contained commits (prefixed `@P0`-`@P3` by priority), each green on its own. Reviewing commit-by-commit is recommended over the squashed diff.

**End state:** 697 tests passing, typecheck clean, lint 0 errors, library build + Storybook (dev & production) build successfully, and **no deprecated direct dependencies remain**.

### React 19
- Fixed removed-API breakages: `Portal` rewritten on `createPortal`; removed `findDOMNode`/`unstable_renderSubtreeIntoContainer`/`ReactDOM.render` usages; fixed a dead string ref (lookup menu), Fragment-invalid-prop leaks, and keyboard nav in menu-dropdown/menu-picklist.
- Removed 165 now-unnecessary `React` imports (jsx-runtime).

### Toolchain
- **Build/test:** Vite 5 -> 8, `@vitejs/plugin-react` 4 -> 6, `vite-plugin-dts` 3 -> 5, Vitest 1 -> 4 (+coverage), jsdom 24 -> 30, jest-dom 6 -> 7.
- **Lint:** ESLint 8 -> 10 with a full flat-config migration; `typescript-eslint` 7 -> 8; moved to the eslint-10-compatible `eslint-plugin-jsx-a11y-x` / `eslint-plugin-import-x` forks (re-keyed to legacy namespaces so inline disables keep working).
- **TypeScript 5 -> 7, side-by-side** per [Microsoft's guidance](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/): TS 7 has no programmatic API until 7.1, so `@typescript/native` (=typescript@7) drives `tsc` while `typescript` resolves to `@typescript/typescript6` for API consumers (typescript-eslint, vite-plugin-dts, editor).
- **Other:** Babel 7 -> 8, nanoid 5 -> 6, `@types/node` -> 24, Node engine floor -> 24, Storybook 10.5.10.

### Testing
- Migrated all 54 Enzyme/Mocha suites to Vitest + RTL; removed the dead Karma/Storyshots pipeline. Re-enabled data-table keyboard-nav, split-view multi-select, and app-launcher highlight tests.

### TypeScript conversion
- Incrementally converted a large share of components and utilities from `.jsx` to `.tsx`, removing dozens of stale ambient `.d.ts` sidecars that shadowed real sources.

### Deprecated dependency removal
- `popper.js` (v1) -> `@floating-ui/dom` (adapter preserves the dialog helpers' expected data shape).
- `lodash.isequal` -> `dequal`.

### Storybook / component fixes
- **column-resizer:** interop shim + alias for the Rolldown UMD-default regression under Vite 8 (`ColumnResizer is not a constructor`).
- **docgen displayName:** `setDisplayName: false` so `react-docgen-typescript` stops clobbering static `displayName` (was breaking GlobalNavigationBar's child filtering, rendering an empty bar); GlobalNavigationBar stories rewritten to use real subcomponents.
- **expression:** default `events` to `{}` so `ExpressionCondition` renders without it.
- **carousel:** replaced dead `via.placeholder.com` URLs with bundled SLDS assets.
- **toast:** close icon sized `medium` (was oversized `large`) to match the notify pattern.
- SLDS 2 CSS served from the npm package; SLDS 2 theme + dark-mode toolbar in Storybook.

### CI
- Real CI workflow; fixed the publish workflow and package artifact; Storybook published to GitHub Pages on push to master.

## Deferred (documented)
- **Vitest 5** - held at 4.1.11 (stable not yet available in the build environment).
- **react-highlighter-ts** - no React 19-compatible release; wrapper casts around the stale types.

## Test plan
- `npm run typecheck` - clean (TS 7)
- `npm test` - 697 passing
- `npm run lint` - 0 errors
- `npm run build` and `npm run build-storybook` - succeed
- GlobalNavigationBar / Carousel / Toast stories verified rendering in a headless browser
---


interactivellama:
First thank you @Niznikr for taking over the branch from Shelby @summerbee !

I'm going to start at a high level based on that I noticed in this branch versus the master branch. 

I see the `2026-07-15-modernization-audit-and-roadmap.md` from July. Is there any more recent analysis? I can query with Claude of course, but I wanted to double check in base I missed something.

## Bugs needing to be fixed
* Icons too large in Combobox menus
* Setup Assistant is broken (we may be able to remove Setup Assistant for see "Needs further discussion with UX Design")
* Missing icons in VisualPicker, Files

## Proposals
- [x] With all these changes the pull request template checklists needs to updated to reflect changes. This should help LLM-assisted bugfix PRs from others.
- [x] NPM package contents: The file names are different. Do all the same imports work? (See below) The original NPM package had different files including the entire source code, since the primary way of consuming the components was from the source code and not from a built package. I'm not sure why this PR started to pivot from that. The primary audits completed appeared to be "matching SLDS2" and not on backwards compatibility. The existing imports, can be found at https://design-system-react-site.herokuapp.com/components/accordions/
  - `import Accordion from '@salesforce/design-system-react/components/accordion';  `
  - `import AccordionPanel from '@salesforce/design-system-react/components/accordion/panel'; `
- [x] We can delete the following components, `menu-picklist`, `notification`, `time-picker`, `lookup` as they have been deprecated for a _very_ long time. (I've likely had deprecation console warning since 2018.)
- [x] HTML Snapshots (via Jest) have been removed. They need to be added back via `.toMatchSnapshot()` for all Storybook stories (HTML and not DOM). This should add to the test coverage, too, as the original Mocha tests did not text renders closely and relied on the Jest snapshots to test the markup. Bonus: This would also be a good way to diff the render output before and after the uplift. 
  - https://github.com/salesforce/design-system-react/blob/master/tests/snapshot-helpers.jsx
  - https://github.com/salesforce/design-system-react/blob/master/tests/story-based-tests.dom-snapshot-test.js
  - Here is the Storyshots to Vitest migration file: https://storybook.js.org/docs/8/writing-tests/snapshot-testing/storyshots-migration-guide#configure-the-testing-framework-for-portable-stories
- [x] Shelby's `2026-08-07-slds2-vs-react-comparison` compared the SLDS2 storybook with DSR's, but the more important task not confirming to the all SLDS2 stories as they can be added later, but confirming that there minimal breaking changes in this library.
- [x] I'm concerned we only have JSDOM-based tests. The original library used Karma so that real browser tests could be performed. Now 11 tests are skipped (see `/migration-discovered-bugs.md`). I propose that we install Vitest+Playwright for these tests? I'm hesitant to force Storybook interactions tests as this is open source, but I'm open to the idea if you think it would be better.
- [x] TypeScript: I can see that the library was converted to TS, but is there a runtime check for non-TS consumers? Something like `babel-plugin-typescript-to-proptypes` comes to mind. For consumers on plain JS/Babel (no TS), they lose the dev-mode console warnings React+PropTypes used to emit for wrong prop types/missing required props at runtime. TS types are erased at build time, so there's zero runtime enforcement for those users now.
- [ ] Docs: I'm open to Storybook as the doc site–however will want to migrate some of the overview docs from the doc website. If we want to keep the docs website, there's no `build:docs` script now. With the update to TS, `react-docgen` should still work, but I don't see a `build:docs` script in `package.json` anymore. `prepare-components` is also missing which builds the site-stories list for the documentation site. It's very concerning that scripts were deleted without it being called out. Package.json's `components` has been removed. This is used by the public website to determine what to show? Have we tested the output with `https://github.com/salesforce-ux/design-system-react-site`?

## Needs further discussion with UX Design
- [ ] Consider deleting BrandBand, AppLauncher, Expression, GlobalHeader, GlobalNavigationBar, LocationMap
- [ ] Possibly missing from SLDS2: SetupAssistant, SplitView. We should run supported components by UXF design

## v0.10.65 NPM package contents
The following is a list of the contents from the previous package. The new package only has type script definitions in the components folder.
- assets
- components
  - `accordion/index.jsx`
- design-system-react-components.js
- design-system-react-components.js.map
- design-system-react-components.min.js
- design-system-react-components.min.js.LICENSE.txt
- design-system-react-components.min.js.map
- design-system-react.js
- design-system-react.js.map
- design-system-react.min.js
- design-system-react.min.js.LICENSE.txt
- design-system-react.min.js.map
- docs
- icons
- lib
- LICENSE.txt
- module
- package.json
- README.md
- styles
- utilities

interactivellama:
@Niznikr I think a nexus NPM package slipped into the packge lock file

interactivellama:
I've modified DeepScan so that only High issues block fail the check above. This should unblock the commits going forward.

interactivellama:
* Setup Assistant should have interactive items on the right and large icons on the left. 

Example from https://design-system-react-site.herokuapp.com/components/setup-assistant/
Screenshot 2026-09-15 at 1 35 04 PM

interactivellama:
## Component and Prop Deprecation
- [x] Illustration: The TypeScript re-write removed the check-props for Illustration. Which is deprecated due to DOM/SVG XSS security issues. Illustration is not in Storybook, but is still in the export/dist.

- [x] Fully removed already (in 9eaff5645, so nothing left to clean up): `menu-picklist`, `notification`, `time-picker`, `lookup`.

- [x]  Still present as "moved" compatibility shims *(we should remove these)*:
  - components/bread-crumb → use components/breadcrumb
  - components/navigation → use components/vertical-navigation
  - components/popover-tooltip → use components/tooltip
  - components/forms/textarea → use components/textarea
  - components/forms/input → use components/input
  - components/forms/radio → use components/radio

 - [ ] Still present but only partially/prop-level deprecated (*we should remove these:* whole component stays, one variant or prop is flagged via componentIsDeprecated):
    - Modal.Trigger (components/modal/trigger.tsx) — deprecated trigger pattern
    - AppLauncher.Section (components/app-launcher/check-props.js) — use AppLauncherExpandableSection instead
    - Illustration (components/illustration/check-props.js) — deprecated due to shadow DOM/SVG security issues
    - GlobalHeader (main, non-subcomponent usage) — deprecated in favor of the specific global-header subcomponents
    - Input.Inline pattern (components/forms/input/inline-check-props.js) — deprecated in favor of the Record-Detail pattern or Popover+Button edit pattern

interactivellama:
UX Foundations design stakeholders via Slack have approved full removal of the following components `BrandBand`, `AppLauncher`, `Expression`, `GlobalHeader`, `GlobalNavigationBar`, and `LocationMap`. This aligns with them not being a supported pattern in SLDS2 and Design System React is an SLDS2-only library after this update. Please remove.

cc: @cliffseal

interactivellama (approved):

